PR #22, opened 2020-08-21
Enable depth aov output

trebconnell:
I'm not totally sure why storm's output is white.  I suspect it's because a different code path is traveled for render because it's renderBuffer is backed by GPU memory, but haven't verified this.  

They both output into depth aov with a single float component.

The results looks equivalent otherwise, and the output from cycles is what I would expect.

Storm depth:
image
Cycles depth:
image

### DIFF
--- a/plugin/hdCycles/renderPass.cpp
+++ b/plugin/hdCycles/renderPass.cpp
@@ -118,7 +118,7 @@ HdCyclesRenderPass::_Execute(HdRenderPassStateSharedPtr const& renderPassState,
         if (numPixels != oldNumPixels) {
             m_colorBuffer.resize(numPixels * pixelSize);
             memset(m_colorBuffer.data(), 0, numPixels * pixelSize);
-            m_depthBuffer.resize(numPixels * pixelSize);
+            m_depthBuffer.resize(numPixels);
         }
     }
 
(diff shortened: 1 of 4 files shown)


plugin/hdCycles/renderPass.cpp
```diff
@@ -144,6 +145,32 @@ HdCyclesRenderPass::_Execute(HdRenderPassStateSharedPtr const& renderPassState,
             if (aov.aovName == HdAovTokens->color) {
                 rb->Blit(colorFormat, w, h, 0, w,
                          reinterpret_cast<uint8_t*>(m_colorBuffer.data()));
+            } else if (aov.aovName == HdAovTokens->depth
+                       || aov.aovName == HdAovTokens->cameraDepth) {
+                renderParam->GetCyclesSession()->buffers->copy_from_device();
+                if (renderParam->GetCyclesSession()->buffers->get_pass_rect(
```
Vochsel, line 151: I get a consistent `Access violation writing location` on this line, when building with USD-19.11 and the latest Cycles standalone build.

Vochsel:
Any updates here @trebconnell? We are about to start work on regular AOV support for our needs. As a heads up it might touch some of the code you've already started updating.

trebconnell:
Got pulled into some other work, happy to hear y'all are doing work here.  Can base off of this if helpful ofc, or start fresh : )
I'll close this.

PR #25, opened 2020-08-25
Uvs with interpolation = "vertex" don't render correctly

trebconnell:
Face varying triangulation is run even for vertex varying uvs.

Images from a sample scene where the plane uses vertex varying uvs.  These are Maya default primitive shapes.

Before:
image

After:
image

### DIFF
--- a/plugin/hdCycles/mesh.cpp
+++ b/plugin/hdCycles/mesh.cpp
@@ -500,6 +500,7 @@ HdCyclesMesh::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam,
 {
     HdCyclesRenderParam* param = (HdCyclesRenderParam*)renderParam;
     ccl::Scene* scene          = param->GetCyclesScene();
+    
 
     scene->mutex.lock();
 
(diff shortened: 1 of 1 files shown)


plugin/hdCycles/mesh.cpp
```diff
@@ -766,19 +767,17 @@ HdCyclesMesh::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam,
                                  ++i) {
```
sherholz, line 767: Despite the fix, it seems that `uvIdndices` is instantiated and filled but never used.
Is there a reason for that or is this still leftover code?

sherholz:
@trebconnell nice that you already added a PR. I just wanted to open an issue for that. :)

trebconnell:
@sherholz Good point.  Removed that code.

sherholz:
@trebconnell I tested your patch with my simple test case and it does not fix the issue:

Screenshot from 2020-08-31 18-57-48

Screenshot from 2020-08-31 18-57-41


The [test case](https://drive.google.com/file/d/1tO3MUl9B6ww5UMm4Y1-c4LpgU4Z0Q63P/view?usp=sharing) is a simple rectangle made out of two triangles with vertex varying uvs:

 ```
    def Mesh "card"
    {
        float3[] extent = [(-430, -145, 0), (430, 145, 0)]
        int[] faceVertexCounts = [3, 3]
        int[] faceVertexIndices = [0, 1, 2, 2, 3, 0]
        rel material:binding = </TexModel/boardMat>
        point3f[] points = [(-430, -145, 0), (430, -145, 0), (430, 145, 0), (-430, 145, 0)]
        texCoord2f[] primvars:st = [(0, 0), (1, 0), (1, 1), (0, 1)] (
            interpolation = "varying"
        )
    }
```

Could you post the USD for your test scene (only containing the rectangle)?

trebconnell:
hdCycles doesn't currently support "varying" interpolation.  If you switch to interpolation = "vertex" it should work.  Updating title to be more explicit here.

"varying" shouldn't be hard to support, but would do in a separate PR.

plugin/hdCycles/mesh.cpp
Vochsel: That was left over code as far as I can remember.

Vochsel (approved):
Can confirm this fixes texcoords for vertex varying, and faceVarying works the same. Thanks @trebconnell

PR #47, opened 2021-01-21
Added compatibility for enums stored as tokens

Vochsel:
For a brief period our houdini_cycles stored enums as tokens instead of strings. This allows support for these usd files.

There is no harm in adding this now, they function similarly enough for now, and this addition will only effect enums anyway,


boberfly:
Hey @Vochsel this LGTM, the only thing now that you've brought up the case break, I think a `default:` on the end is needed to keep GCC happy and we can use `default:` to print out anything which passes through and doesn't convert so we are aware of things passing through. What do you think? Could be another PR though.

Vochsel:
@boberfly  That's a good point. I'll add it on the end of this one. Thanks!

PR #50, opened 2021-01-25
Improvement: Arbitrary mesh attributes

Vochsel:
This PR is a step towards a more generic way of populating Cycles attributes from meshes.

This does not touch curves or other attributes. It eventually should be agnostic to all.

### DIFF
--- a/plugin/hdCycles/utils.cpp
+++ b/plugin/hdCycles/utils.cpp
@@ -550,101 +550,101 @@ HdCyclesIsPrimvarExists(TfToken const& a_name,
 }
 
 template<>
-float
-to_cycles<float>(const float& v)
+inline float
+to_cycles<float>(const float& v) noexcept
 {
     return v;
 }
 template<>
-float
-to_cycles<double>(const double& v)
+inline float
+to_cycles<double>(const double& v) noexcept
 {
     return static_cast<float>(v);
 }
 template<>
-float
-to_cycles<int>(const int& v)
+inline float
+to_cycles<int>(const int& v) noexcept
 {
     return static_cast<float>(v);
 }
 
 
 template<>
-ccl::float2
-to_cycles<GfVec2f>(const GfVec2f& v)
+inline ccl::float2
+to_cycles<GfVec2f>(const GfVec2f& v) noexcept
 {
     return ccl::make_float2(v[0], v[1]);
 }
 template<>
-ccl::float2
-to_cycles<GfVec2h>(const GfVec2h& v)
+inline ccl::float2
+to_cycles<GfVec2h>(const GfVec2h& v) noexcept
 {
     return ccl::make_float2(static_cast<float>(v[0]), static_cast<float>(v[1]));
 }
 template<>
-ccl::float2
-to_cycles<GfVec2d>(const GfVec2d& v)
+inline ccl::float2
+to_cycles<GfVec2d>(const GfVec2d& v) noexcept
 {
     return ccl::make_float2(static_cast<float>(v[0]), static_cast<float>(v[1]));
 }
 template<>
-ccl::float2
-to_cycles<GfVec2i>(const GfVec2i& v)
+inline ccl::float2
+to_cycles<GfVec2i>(const GfVec2i& v) noexcept
 {
     return ccl::make_float2(static_cast<float>(v[0]), static_cast<float>(v[1]));
 }
 
 template<>
-ccl::float3
-to_cycles<GfVec3f>(const GfVec3f& v)
+inline ccl::float3
+to_cycles<GfVec3f>(const GfVec3f& v) noexcept
 {
     return ccl::make_float3(v[0], v[1], v[2]);
 }
 template<>
-ccl::float3
-to_cycles<GfVec3h>(const GfVec3h& v)
+inline ccl::float3
+to_cycles<GfVec3h>(const GfVec3h& v) noexcept
 {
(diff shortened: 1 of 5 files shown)


plugin/hdCycles/utils.cpp
```diff
@@ -474,6 +526,342 @@ HdCyclesIsPrimvarExists(TfToken const& a_name,
...
+        data_size = sizeof(ccl::float2);
+    else if (num_components == 3)
+        data_size = sizeof(ccl::float3);
+    else if (num_components == 4)
+        data_size = sizeof(ccl::float4);
+
+    if (interpolation == HdInterpolationVertex) {
+        VtIntArray::const_iterator idxIt = mesh->GetFaceVertexIndices().begin();
+
+        for (int i = 0; i < value.GetArraySize(); i++) {
```
boberfly, line 628: Like we kind of talked about, the heavy branch checking for every value in the array will be bad for performance.

Possibly what we could do here is re-arrange the loops so that the top of the check is the `value.IsHolding<VtArray<T>>()` (meaning for all the permutations above where you work out num_components/src_type), followed by a check for what interpolation it is each one, then do the for-loop for each of those, so all the branchy is out of the way once we hit the loop. If that makes sense?

plugin/hdCycles/utils.cpp
```diff
@@ -474,6 +526,342 @@ HdCyclesIsPrimvarExists(TfToken const& a_name,
...
+    VtVec2dArray colorVec2d;
+    VtVec3dArray colorVec3d;
+    VtVec4dArray colorVec4d;
+
+    VtIntArray colorVec1i;
+    VtVec2iArray colorVec2i;
+    VtVec3iArray colorVec3i;
+    VtVec4iArray colorVec4i;
+
+    // 0 float, 1 double, 2 int
```
bareya, line 555: Strong enum would be nice. It's hard to read the code if we have to lookup what type is under what index.

plugin/hdCycles/utils.cpp
```diff
@@ -474,6 +526,342 @@ HdCyclesIsPrimvarExists(TfToken const& a_name,
...
+    int src_type = 0;
+
+    if (value.IsHolding<VtArray<float>>()) {
+        colorVec1f     = value.UncheckedGet<VtArray<float>>();
+        num_components = 1;
```
bareya, line 560: I checked the docs, all `GfVec*` classes carry `static const size_t dimension`. That means everything here can be wrapped very nicely in a template function. With type src_type being deduced based on presence of an alias.

plugin/hdCycles/utils.cpp
```diff
@@ -298,6 +300,18 @@ vec2f_to_float2(const GfVec2f& a_vec)
     return ccl::make_float2(a_vec[0], a_vec[1]);
 }
 
+ccl::float2
+vec2i_to_float2(const GfVec2i& a_vec)
```
bareya, line 304: I would move away from naming functions this way and push them into a header file to allow compiler to inline them when necessary. I'd rely on overload resolution:

```c++
inline ccl::float2 to_float2(const GfVec2i& v) noexcept {
    return  ccl::make_float2(static_cast<float>(v[0]), static_cast<float>(v[1]));
}

inline ccl::float2 to_float2(const GfVec2f& v) noexcept {
    return  ccl::make_float2(v[0], v[1]);
}

File: plugin/hdCycles/utils.cpp [hunk 5]
```diff
@@ -499,6 +550,342 @@ HdCyclesIsPrimvarExists(TfToken const& a_name,
...
+                int v0 = count;
+                int v1 = (count + j + 0);
+                int v2 = (count + j + 1);
+
+                if (mesh->GetOrientation() == HdTokens->leftHanded) {
+                    v1 = (count + ((vCount - 1) - j) + 0);
+                    v2 = (count + ((vCount - 1) - j) + 1);
+                }
+
+                if (num_components == 1) {
```
bareya, line 757: There is a lot of code duplication here, can't we wrap it into a function?

plugin/hdCycles/utils.cpp
Vochsel: This has been done, the only if statement we have in a for loop for attributes is to check orientation, which @bareya will likely solve with his proposed work on winding order and Osd.

plugin/hdCycles/utils.cpp
Vochsel: No longer needed

plugin/hdCycles/utils.cpp
Vochsel: Unfortunately, vec1's or just plain old float, double, and int don't have these and mean its not as nice to solve. However in the latest revision, I went with a templated `to_cycles` function that is deduced based on the needs from cycles. 
It's probably possible to go more down the template rabbit hole, however this feels like good enough for now, considering the conversations in the last couple days about re-doing this code again with the findings of @bareya

plugin/hdCycles/utils.cpp
Vochsel: Addressed in latest revision

plugin/hdCycles/utils.cpp
Vochsel: We can overhaul and refactor the rest of these to your suggestion, but lets do that in another PR. for now, I have added the `to_cycles` templated function which may or may not help the situation

PR #51, opened 2021-02-02
Support for FaceVarying and Uniform normals

dedoardo:
Adding support for mesh normals with FaceVarying and Uniform interpolation by setting them as corner attributes in the Cycles mesh.  This [PR](https://github.com/tangent-opensource/cycles/pull/3) is required.

Allowing normals to be specified on subdivision surfaces (Cycles side) requires filling in different arrays. I don't know if we want to pursue this direction or adhere to the USD guidelines more strictly.


plugin/hdCycles/mesh.cpp
```diff
@@ -438,19 +438,29 @@ HdCyclesMesh::_AddNormals(VtVec3fArray& normals, HdInterpolation interpolation)
...
+        ccl::Attribute* attr_fvN = attributes.add(ccl::ATTR_STD_CORNER_NORMAL);
+        ccl::float3* fvN         = attr_fvN->data_float3();
 
         int idx = 0;
         for (int i = 0; i < m_faceVertexCounts.size(); i++) {
             const int vCount = m_faceVertexCounts[i];
+            ccl::float3 fNi = vec3f_to_float3(normals[i]);
+            if (m_orientation == HdTokens->leftHanded) {
+                fNi = -fNi;
+            }
```
Vochsel, line 456: A profiler may show more info, and this might be redundant, but perhaps setting a scale outside of this for loop with direction, and then multiplying all conversions by the scale might be faster? 

```
float scale = (m_orientation == HdTokens->leftHanded) ? -1.0f : 1.0f;

for (int i = 0; i < m_faceVertexCounts.size(); i++) {
            const int vCount = m_faceVertexCounts[i];	 
            ccl::float3 fNi = vec3f_to_float3(normals[i]) * scale;
            
```

Vochsel: This is more me thinking out loud.

plugin/hdCycles/mesh.cpp
```diff
@@ -438,19 +438,29 @@ HdCyclesMesh::_AddNormals(VtVec3fArray& normals, HdInterpolation interpolation)
...
     if (interpolation == HdInterpolationUniform) {
-        ccl::Attribute* attr_fN = attributes.add(ccl::ATTR_STD_FACE_NORMAL);
-        ccl::float3* fN         = attr_fN->data_float3();
+        // Expecting one normal for each polygon
+        if (normals.size() != m_faceVertexCounts.size()) {
+            TF_WARN("Unexpected number of uniform normals, skipping.");
+            return;
+        }
+
+        ccl::Attribute* attr_fvN = attributes.add(ccl::ATTR_STD_CORNER_NORMAL);
```
Vochsel, line 447: Can't we use ccl::ATTR_STD_FACE_NORMAL, and save us the data?

plugin/hdCycles/mesh.cpp
```diff
@@ -438,19 +438,29 @@ HdCyclesMesh::_AddNormals(VtVec3fArray& normals, HdInterpolation interpolation)
...
+                fNi = -fNi;
+            }
 
-            // This needs to be checked
-            for (int j = 1; j < vCount - 1; ++idx) {
-                fN[idx] = vec3f_to_float3(normals[idx]);
+            for (int j = 1; j < vCount - 1; ++j) { // for each triangle
+                for (int k = 0; k < 3; ++k) { // for each corner
+                    fvN[idx++] = fNi;
+                }
```
Vochsel, line 461: if we can use `ccl::ATTR_STD_FACE_NORMAL` this becomes less data

plugin/hdCycles/mesh.cpp
```diff
@@ -818,6 +818,8 @@ HdCyclesMesh::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam,
                                               id, HdInterpolationFaceVarying) },
             { HdInterpolationVertex,
               sceneDelegate->GetPrimvarDescriptors(id, HdInterpolationVertex) },
+            { HdInterpolationUniform,
+              sceneDelegate->GetPrimvarDescriptors(id, HdInterpolationUniform) },
```
Vochsel, line 822: I also added this in my Attribute PR, should be noted when merging these PRS

plugin/hdCycles/mesh.cpp
```diff
@@ -465,34 +475,24 @@ HdCyclesMesh::_AddNormals(VtVec3fArray& normals, HdInterpolation interpolation)
...
+                fvN[i * 3 + j] = vec3f_to_float3(normals[i * 3 + j]);
 
-        // For now, we add all corner normals and normalize separately.
-        // TODO: Update when Cycles supports corner_normals
-        /*for (size_t i = 0; i < m_numMeshFaces; i++) {
-            for (size_t j = 0; j < 3; j++) {
-                ccl::float3 n = vec3f_to_float3(normals[(i * 3) + j]);
-                cdata[m_cyclesMesh->get_triangle(i).v[j]] += n;
+                if (m_orientation == HdTokens->leftHanded) {
+                    fvN[i * 3 + j] = -fvN[i * 3 + j];
```
Vochsel, line 492: Again, maybe we can save the branching and just multiply by a scale, but I also have not tested the performance of this.

plugin/hdCycles/mesh.cpp
Vochsel: @boberfly @bareya might have done this before

Vochsel (changes requested):
Overall looks good. There is an issue with leftHanded vertex normals but this will hopefully be solved with @bareya's work. 

Please also format this with a clang formatter if possible.

If we can use ccl::ATTR_STD_FACE_NORMAL for HdInterpolationUniform, I would have thought that preferrable, but I might be missing something on the cycles side.

The notes about performance and moving leftHanded branch to a multiplication is me thinking out loud. These changes are not necessary imo.  

Required changes:
 * formatting
 * maybe switch to face normals
 * maybe remove if statement from for loops and use a multiplication instead

plugin/hdCycles/mesh.cpp
boberfly: I think in cortex/gaffer they seemed to just match so I didn't need to deal with handedness (USD scene reader seems to fix it up on-read).

bareya: @Vochsel you are right, branching should be pull out from the for loop. This is how it's done in PxOsd.

boberfly:
I wanted to add here that MikkTspace stuff will need looking at also for hdCycles (see note of it in the cycles PR):
https://github.com/tangent-opensource/cycles/pull/3

I think if a normal map is used on faceVarying normals it'll crash currently, so it'll need looking into to do the right thing when creating tangents/bi-tangents.

dedoardo:
Addressed the formatting and the flipping outside the loop. 

Regarding custom face normals, unless I am totally wrong, Cycles does not really support them. [The geometric normal is computed when the ray hits the surface using the triangle points](https://github.com/tangent-opensource/cycles/blob/master/src/kernel/kernel_shader.h#L95) and used as shading normal unless vertex normals are present. Essentially, it won't do anything with `ATTR_STD_FACE_NORMAL`. 

The good news is that corner normals in Cycles put the machinery in place to have different types of normals per object. This requires deciding how to expose which types of normal are to be used because now we would have 3 different types rather than 2. 
An order of precedence would work but it's a bit hidden and could run into issues as the geometry code is keen on recomputing normals. @skwerner do you have any thoughts on what's a reasonable way to proceed?

Once that is discussed it shouldn't be too hard to include it in Cycles in the same corner PR or have a separate one for it.

dedoardo:
@boberfly The tangent space calculation is a good point. I don't think it will crash, but will produce wrong results as it will use the vertex normals rather than the custom ones (corner or face). 
It can be addressed, I need to create a scene with normal mapping and see what happens.

plugin/hdCycles/mesh.cpp
Vochsel: Thanks for the clarification about this. If @skwerner has no other thoughts, this seems like a reasonable solution.

Vochsel (approved):
Code here looks good. Strange situation with ATTR_STD_FACE_NORMAL, but I'll take your word for it, your solution seems good. Will the tangent fix @boberfly mentioned will be part of this PR or another? If not, this looks good to approve, but is held by the Cycles PR

plugin/hdCycles/mesh.cpp
Vochsel: Thanks for this fix in the latest push @dedoardo :)

dedoardo:
The mikktspace callbacks now use corner normals over vertex/face when present. The code that computes the tangents has been moved out of the attributes loop to guarantee that the correct normals are available.

I tested it on a couple of meshes with normal mapping and it seems to not break anything.

PR #54, opened 2021-02-03
remove pxr namespace

bareya:
If USD is compiled without name spaces. `hdCycles` will not compile since `pxr` namespace does not exist. We don't have to use `pxr::` namespace because we are in between `PXR_NAMESPACE_OPEN_SCOPE` `PXR_NAMESPACE_CLOSE_SCOPE`


Vochsel (approved):
Yep, fine by me.

PR #56, opened 2021-02-03
remove USD_HAS_NEW_SDR_NODE_CONSTRUCTOR define

bareya:
we do not need `USD_HAS_NEW_SDR_NODE_CONSTRUCTOR` we can use `PXR_` macros explicitly


Vochsel:
We can also remove from here: https://github.com/tangent-opensource/hdcycles/blob/main/CMakeLists.txt#L220

Vochsel:
Happy with this change.

PR #58, opened 2021-02-04
Improved stability for render settings node

Vochsel:
Updated description:
The main problem solved here is stopping the render pass blitting when the render buffer has just been deallocated. 

Along with this, we are also removing the mutex locks for all primitive adding and deleting in the render param. These are no longer needed.
 
Testing has been done and stability has increased.


boberfly:
With this patch it also removes all those scene mutex locks, I'm assuming you were experiencing crashes and put them all in place to prevent it, but then later discovered the crashes were due to Houdini clearing the render buffer? Sorry I'm just trying to understand this patch better as the description isn't very informative to what the patch is doing per se.

Vochsel:
yep, the mutex locks were added early on. It no makes more sense to lock the mutex outside of these. I.e. when adding a mesh and its object, we lock around the higher level calls. 

From my tests, removing these causes no issues in an interactive viewport.

Vochsel:
On top of this, I also think they were causing render failures on the farm, but I have less evidence of this until we test on farm

bareya (approved):

bareya:
In general, for the future, instead of using `lock()`/`unlock()` mechanism, that eventually leads to errors, it's better to use some sort of scoped lock that does it automatically for you.

Vochsel:
Agreed, @bareya unfortunately cycles doesnt offer too much in that department. they offer try_lock, but in my original tests, this did not work well with Hydra.

bareya:
@Vochsel in `util_thread.h` you can find aliases:

```c++
typedef std::mutex thread_mutex;
typedef std::unique_lock<std::mutex> thread_scoped_lock;
```
Seems like you can use `thread_scoped_lock`, or `std::lock_guard` next time.

Issue #62, opened 2021-02-04
Support for UsdLuxShadowAPI


Vochsel:
https://graphics.pixar.com/usd/docs/api/class_usd_lux_shadow_a_p_i.html

We will likely need to remove these options from the usdCycles api to avoid duplicates...

PR #63, opened 2021-02-05
Switched from custom usdCycles schema to usdlux shadow schema

Vochsel:
As per https://github.com/tangent-opensource/hdcycles/issues/62

We now use the UsdLux schema to determine if shadows should be cast from a light.

Tested in an interactive Solaris session


boberfly:
LGTM!

boberfly (approved):
LGTM!

PR #67, opened 2021-02-05
Mesh refactoring and adding Osd uniform subdivision

bareya:
## Osd topology and PrimVars refinement

Screenshot from 2021-02-14 15-01-32

_( 1.Triangulated input quadrilaterals 2. Osd catmull clark scheme 3. Osd catmul clark scheme with boundary set to edge only)_

* Right, Left handed topology is converted to right handed.
* Topology and attributes are refined through `OpenSubdiv` or `MeshUtil`
* No need for custom schema for subdivision, but refinement level for individual object has to be handled through `usdCycles` schema.
* One interface for refinement, clear and separate implementation for triangulation and subdivision. 
* Multi threaded geometry conversion and refinement.
* Geometry uses resize rather than reserve for stability reasons.
* No support for holes in Osd topology 
* Tbb as default PrimVars evaluator

## Creases

Screenshot from 2021-02-25 14-17-07

## Limit Surface normals evaluation

limit_surface_normals

When geometry is subdivided, then normals are generated automatically.

## Face Varying UV sets with linear interpolation

Screenshot from 2021-02-25 09-40-54

_(1. Input mesh, 2. Linear interpolation, 3. Smooth interpolation)_

Currently Hydra does not support custom topologies for face varying primvars. Each patch is treated as independent island. Therefore number 3 has stretched uvs.

Screenshot from 2021-02-25 10-01-22

## Materials

Screenshot from 2021-02-16 09-03-20

_(GeomSubsets - red, green, blue. default material - top row)_

* Removed fallback material, and introducing default material.
* Object level material overrides default material
* GeomSubset appends materials and overrides material id per triangle
* MaterialId change triggers topology rebuild.
* Hash based lookup for material ids.

## Color attribute

Screenshot from 2021-02-25 09-50-57

_(1. Constant, 2. Uniform, 3. Vertex, 4. Varying, 5. Linear Face Varying, 6. Smooth Face Varying )_

* Introduced default materials to read Constant, Uniform, Vertex and Varying primvars. Kitchen Scene is renders with colors now.
* Fixed incorrect input data conversion to non float3(color) data.
* Fixed Varying attribute conversion.

## Kitchen Scene(no instancing)

Screenshot from 2021-02-25 09-32-31
_(Uniform Subdivision set to high)_

Current implementation of instances, used to crash and still crashes Kitchen Scene


plugin/hdCycles/mesh.cpp
```diff
@@ -557,96 +663,295 @@ HdCyclesMesh::_PopulateMotion()
...
+
+    constexpr int default_shader_id = 0;
+    for(size_t i{}; i < m_cyclesMesh->num_triangles(); ++i) {
+        m_cyclesMesh->shader[i] = default_shader_id;
+    }
+
+    _PopulateObjectMaterial(sceneDelegate, scene, id);
+    _PopulateSubSetsMaterials(sceneDelegate, scene, id);
+
+    static std::mutex material_mutex;
```
dedoardo, line 727: Why static and not a member?

dedoardo (changes requested):
Really great job all round, both refactoring and supporting all the features!!

From a couple of tests it seems to work. I left some minor comments around, I am not too familiar with the OpenSubdiv API, maybe someone else can give better advice on that end, but as long as nothing else is broken, we should look to merge this soon.

Two extra comments

- _PopulateMotion needs to be called somewhere
- Honestly, I do not understand the behavior of the color attribute in Column 3 (4 a bit more). Why is the bottom patch so different than the top one in the central 4 squares? I'm guessing that the non-squared boundary mean that they are creases? Still, the interpolation seems a bit weird... Although it's true that you can't see the per-vertex values.

plugin/hdCycles/mesh.cpp
bareya: This is a left over from me trying to find out why the code was crashing. Definitely that mutex should be a member. Good find!

dedoardo:
I have merged dev into the current branch, things seem to work fine except for deformation blur which behaves erratically and crashes, it's thus commented out right now. 
Found two bugs along the way
- The interpolation type for the normals was not initialized/set and resulted in normals not being read.
- There was a mismatch between the normal interpolation type and the refiner method that was being called. The implementations look the same, so I don't think this was causing any issues.

plugin/hdCycles/mesh.cpp
```diff
@@ -633,6 +633,7 @@ HdCyclesMesh::_PopulateTopology(HdSceneDelegate* sceneDelegate, const SdfPath& i
     auto refine_value         = GetPrimvar(sceneDelegate, usdCyclesTokens->primvarsCyclesMeshSubdivision_max_level);
     int refine_level          = refine_value.IsEmpty() ? 0 : refine_value.Cast<int>().UncheckedGet<int>();
     display_style.refineLevel = refine_level;
+    display_style.refineLevel = 5;
```
bareya: Has this been removed?

plugin/hdCycles/mesh.cpp
```diff
@@ -1192,14 +1192,16 @@ HdCyclesMesh::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam, H
 
     if (*dirtyBits & HdChangeTracker::DirtyVisibility) {
         _sharedData.visible = sceneDelegate->GetVisible(id);
-        _UpdateObject(scene, param, dirtyBits);
+        _UpdateObject(scene, param, dirtyBits, false);
```
bareya, line 1221: Super cool that you remembered about this one :+1:

bareya:
> * The interpolation type for the normals was not initialized/set and resulted in normals not being read.

@dedoardo Can you point to the code please, I don't see it.

dedoardo:
[Here](https://github.com/tangent-opensource/hdcycles/blob/430134fa5c7b7ae89665cbfaec4d413ceae5c247/plugin/hdCycles/mesh.cpp#L455), the interpolation should be set to the index in the loop, but remains initialized although the function returns true.

.clang-format
```diff
@@ -50,7 +50,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeComma
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: false
-ColumnLimit:     80
+ColumnLimit:     120
```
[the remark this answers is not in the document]
skwerner: This is unrelated to mesh and OSD, I think this would be better as a separate PR.

dedoardo:
I have reverted the behavior of scene locks to match the one in `dev`, unfortunately I don't think we cannot get rid of the scene locks when editing cycles objects but [we should look to be smarter about their usage](https://github.com/tangent-opensource/hdcycles/issues/89). 
Excluding this, I think this PR is pretty much ready to go. We might want to merge [the normal fix](https://github.com/tangent-opensource/hdcycles/pull/88) before or right after merging this PR as it fixed shading issues in time varying geometry.

dedoardo (approved):

plugin/hdCycles/meshRefiner.cpp
```diff
@@ -0,0 +1,682 @@
...
+#include <pxr/imaging/pxOsd/refinerFactory.h>
+
+#include <opensubdiv/far/patchMap.h>
+#include <opensubdiv/far/patchTableFactory.h>
+#include <opensubdiv/far/primvarRefiner.h>
+#include <opensubdiv/far/ptexIndices.h>
+#include <opensubdiv/far/stencilTableFactory.h>
+#include <opensubdiv/osd/cpuPatchTable.h>
+#include <opensubdiv/osd/cpuVertexBuffer.h>
+#include <opensubdiv/osd/tbbEvaluator.h>
```
skwerner, line 36: We don't have tbbEvaluator available in our build variants 0 and 1 (that is, building against our own OSD Rez packages instead of Houdini's).

__has_include() could be an option to use TBB where available (with a cpuEvaluator.h fallback) if we move up to C++17.

For now, I'd suggest switching to cpuEvaulator to get releases out, then revisit TBB as a followup optimisation step.

skwerner (changes requested):

plugin/hdCycles/meshRefiner.cpp
dedoardo: @skwerner True, I should have pointed out that I had to build opensubdiv with TBB support. I think it's a chance that the houdini variants have it. The `rez_deps` file always specifies `NO_TBB = 1` but some package along the way does `FindPackage(TBB)` and the opensubdiv cmake picks it up [because it's checking for `TBB_FOUND`](https://github.com/tangent-opensource/hdcycles/pull/88).

Issue #66, opened 2021-02-05
Default dome light is not handled correctly


Vochsel:
Hydra renderers tend to have a default dome light so that initial stages with no lights present are rendererable (See karma, renderman, etc). Once a light of any kind is added to the scene, the default dome light should go to black (unless an actual dome light is added)

This is partially implemented but at some point, the default dome light is not turning off properly....

PR #69, opened 2021-02-05
Bugfix: Fixed default dome light behaviour

Vochsel:
The default dome light would not disappear after adding lights. This now brings the behaviour in line with other Hydra delegates.

See https://github.com/tangent-opensource/hdcycles/issues/66 for more info

### DIFF
--- a/plugin/hdCycles/renderParam.cpp
+++ b/plugin/hdCycles/renderParam.cpp
@@ -1484,7 +1484,6 @@ void
 HdCyclesRenderParam::CommitResources()
 {
     if (m_shouldUpdate) {
-        std::cout << "m_numDomeLights: " << m_numDomeLights << '\n';
         if (m_cyclesScene->lights.size() > 0) {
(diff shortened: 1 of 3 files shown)


plugin/hdCycles/renderParam.cpp
```diff
@@ -1483,8 +1484,9 @@ void
 HdCyclesRenderParam::CommitResources()
 {
     if (m_shouldUpdate) {
+        std::cout << "m_numDomeLights: " << m_numDomeLights << '\n';
```
bareya, line 1487: Can we switch to `TF_*` macros? I think controlling this kind of logging is just painful.

Vochsel: Oh, thats a debug statement :D I'll remove that. But yes, there are a few places of logging. We can clear that up in another PR.

Vochsel:
Removed that print, let me know if theres any other changes needed @bareya

PR #70, opened 2021-02-05
Bugfix: Added support for Boolean sockets

Vochsel:
Boolean sockets were never supported by HdCycles until now.


boberfly (approved):
LGTM! I'm sure it is working with your test case?

Vochsel:
Yep, although we dont have any as granular as this for the params... We should....

PR #71, opened 2021-02-06
Cleanup: Replaced stdout's with TF_* helpers 

Vochsel:
All cout's of HdCycles have been removed. 

Sidenote: Houdini/Solaris doesnt seem to output TF_STATUS or TF_WARN to the Houdini console. Is there an env var for that? They mention that you can set `TF_DEBUG=PLUG_*` but don't think that works in this case. When run through usdview, the warnings and status' do get written to stdout.

Except for: 

https://github.com/tangent-opensource/hdcycles/blob/improvement/remove_couts/plugin/hdCycles/renderParam.cpp#L148

and

https://github.com/tangent-opensource/hdcycles/blob/improvement/remove_couts/plugin/hdCycles/renderParam.cpp#L165

We do actually want these going into the stdout stream for Deadline to pickup on render progress and logging. @bareya if you have any thoughts, happy to change this.

### DIFF
--- a/plugin/hdCycles/config.cpp
+++ b/plugin/hdCycles/config.cpp
@@ -92,9 +92,6 @@ TF_DEFINE_ENV_SETTING(HD_CYCLES_ENABLE_PROGRESS, false,
 TF_DEFINE_ENV_SETTING(HD_CYCLES_USE_TILED_RENDERING, false,
                       "Use Tiled Rendering (Experimental)");
 
-TF_DEFINE_ENV_SETTING(HD_CYCLES_UP_AXIS, "Z",
-                      "Set custom up axis (Z or Y currently supported)");
-
 // HdCycles Constructor
 HdCyclesConfig::HdCyclesConfig()
 {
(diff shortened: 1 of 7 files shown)


plugin/hdCycles/config.cpp
```diff
@@ -111,7 +108,7 @@ HdCyclesConfig::HdCyclesConfig()
     enable_logging  = TfGetEnvSetting(HD_CYCLES_ENABLE_LOGGING);
     enable_progress = TfGetEnvSetting(HD_CYCLES_ENABLE_PROGRESS);
 
-    up_axis = TfGetEnvSetting(HD_CYCLES_UP_AXIS);
+    up_axis = HdCyclesEnvValue<std::string>("HD_CYCLES_UP_AXIS", "Z");
```
Vochsel, line 111: I also changed the HD_CYCLES_UP_AXIS env var to not be a setting. `TF_DEFINE_ENV_SETTING` output to stdout when overridden and cannot be turned off. This was causing a inconvenience in Solaris with a popup log box every time the render was restarted.

bareya:
> We do actually want these going into the stdout stream for Deadline to pickup on render progress and logging. @bareya if you have any thoughts, happy to change this.

I sent a message to SideFx about this problem and I CCd you. Does it mean that deadline works only with `stdout` and ignores `stdlog` and `stderr`?

PR #75, opened 2021-02-10
Feature/volume motion blur

sergeneren:
#### Summary

Adds support for rendering volumes with deformation motion blur 

#### Change Log 

- Added a new function to *utils.h* to parse usdCyclesSchema API params
- Fixed documentation for HdCyclesVolume and HdCyclesOpenvdbAsset

#### Testing 

- Please build cycles with this branch: [feature/volume-motion-blur](https://github.com/tangent-opensource/cycles/tree/feature/volume-motion-blur)
- If you are testing inside houdini please launch houdini with env var: `$Env:HD_CYCLES_UP_AXIS = "Y"`
- Use the below test files or a custom volume with *vel* field to check volume motion blur. 

[test_files.zip](https://github.com/tangent-opensource/hdcycles/files/5959965/test_files.zip)

### DIFF
--- a/plugin/hdCycles/camera.cpp
+++ b/plugin/hdCycles/camera.cpp
@@ -71,10 +71,10 @@ EvalCameraParam(T* value, const TfToken& paramName,
 
 #ifdef USE_USD_CYCLES_SCHEMA
 
-std::map<TfToken, ccl::Camera::MotionPosition> MOTION_POSITION_CONVERSION = {
-    { usdCyclesTokens->start, ccl::Camera::MOTION_POSITION_START },
-    { usdCyclesTokens->center, ccl::Camera::MOTION_POSITION_CENTER },
-    { usdCyclesTokens->end, ccl::Camera::MOTION_POSITION_END },
+std::map<TfToken, ccl::MotionPosition> MOTION_POSITION_CONVERSION = {
+    { usdCyclesTokens->start, ccl::MOTION_POSITION_START },
+    { usdCyclesTokens->center, ccl::MOTION_POSITION_CENTER },
+    { usdCyclesTokens->end, ccl::MOTION_POSITION_END },
 };
 
 std::map<TfToken, ccl::Camera::RollingShutterType> ROLLING_SHUTTER_TYPE_CONVERSION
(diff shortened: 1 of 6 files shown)


plugin/hdCycles/camera.cpp
```diff
@@ -484,16 +485,17 @@ HdCyclesCamera::ApplyCameraSettings(ccl::Camera* a_camera)
...
-        = ccl::Camera::MotionPosition::MOTION_POSITION_CENTER;
+        = ccl::MotionPosition::MOTION_POSITION_CENTER;
 
     a_camera->rolling_shutter_duration = m_rollingShutterTime;
 
     a_camera->rolling_shutter_type
         = (ccl::Camera::RollingShutterType)m_rollingShutterType;
     a_camera->panorama_type   = (ccl::PanoramaType)m_panoramaType;
-    a_camera->motion_position = (ccl::Camera::MotionPosition)m_motionPosition;
+    a_camera->motion_position = (ccl::MotionPosition)m_motionPosition;
```
bareya, line 498: There are many benefits of using `static_cast<>`, next time when you see C-style casts switch them to `static_cast` please.

plugin/hdCycles/utils.h
```diff
@@ -602,6 +603,30 @@ _HdCyclesGetCameraParam(HdSceneDelegate* a_scene, SdfPath a_id, TfToken a_token,
     return _HdCyclesGetVtValue<T>(v, a_default);
 }
 
+// Get Volume param
+
+template<typename T>
+T
+_HdCyclesGetVolumeParam(const HdPrimvarDescriptor& a_pvd,
```
bareya, line 610: * Isn't this a utility function that was pulled out from somewhere and has been renamed? I am sure I have seen it elsewhere before, HdCyclesMesh probably. We don't want to duplicate the code do we? 

* This function is asking for const correctness review. `a_token` can be const reference, `HdVolume` can be const pointer. 

* `T a_default` should be passed as const reference. We don't want to copy it just to call the function. We want to copy it when we return.

bareya (approved):
Looks good but I would re-visit the code mentioned in the comments.

PR #76, opened 2021-02-16
Uniform steps for transformation blur

dedoardo:
**Summary**
More consistent rendering of motion blur by recalculating transforms at uniform time steps. It adapts to any values returned by `SampleTransform` and the results are matching Karma/RPR.

**Changes**
-  Resampling Mesh transforms during `Sync` using provided samples, always resulting in an odd number of samples.

**Testing**
[Test file](https://github.com/tangent-opensource/hdcycles/files/5990588/test.zip)

**Frame 1 Before/After**
before_1
after_1


plugin/hdCycles/utils.cpp
```diff
@@ -218,18 +201,84 @@ HdCyclesSetTransform(ccl::Object* object, HdSceneDelegate* delegate,
...
+                ccl::Transform xfPrev = mat4d_to_transform(
+                    xf.values.data()[iXfPrev]);
+                ccl::Transform xfNext = mat4d_to_transform(
+                    xf.values.data()[iXfNext]);
+
+                ccl::DecomposedTransform dxf[2];
+                transform_motion_decompose(dxf + 0, &xfPrev, 1);
+                transform_motion_decompose(dxf + 1, &xfNext, 1);
+
+                // Preferring the smaller rotation difference
```
boberfly, line 263: This is more for my own curiosity, does this make sure the quaternion slerp work properly under the hood?

plugin/hdCycles/utils.cpp
```diff
@@ -218,18 +201,84 @@ HdCyclesSetTransform(ccl::Object* object, HdSceneDelegate* delegate,
...
 
-        object->motion.resize(sampleCount, ccl::transform_empty());
+        // Rounding to odd number of samples to have one in the center
+        const int sampleOffset     = (sampleCount % 2) ? 0 : 1;
+        const int numMotionSteps   = sampleCount + sampleOffset;
+        const float motionStepSize = (xf.times.back() - xf.times.front())
+                                     / (numMotionSteps - 1);
+        object->motion.resize(numMotionSteps, ccl::transform_empty());
+
+        // For each step, we use the available data from the neighbors
```
boberfly, line 211: Nice, this correctly unravels non-uniform steps to what Cycles expects/uniform? Going to study this for GafferCycles... :)

boberfly (commented):
Hey @dedoardo this is just more or less questions/statements, the matching motion blur to Karma/RPR is very promising!

plugin/hdCycles/utils.cpp
dedoardo: It's because I noticed that when two rotations with different signs were being interpolated it resulted in the rotation to go in the opposite (longer) direction. `transform_motion_array_interpolate` does not seem to do any check like this, maybe it's done internally or indirectly. I will look into it more?

plugin/hdCycles/utils.cpp
dedoardo: The code tries to make the least number of assumptions, every time Cycles expects a step I look for the closest samples in time and interpolate them or copy them if the times match. 
The only thing that I am puzzled about is that `SampleTransform` guarantess to return 2 samples at the shutter open/close time which are calculated by hydra using linear interpolation. This results in a bit of a weird behavior (see before frame 1 The length of the stick is more correct but the rotation is visually wrong).. It would be ideal to have quaternion interpolation or access to the raw samples to calculate it ourselves.

skwerner (approved):

PR #77, opened 2021-02-23
Enabling motion blur for geometries with varying topology

dedoardo:
**Overview**
Enabling motion blur to be computed from velocities and accelerations.

**Changes**
- Triggering rebuild of the BVH for meshes with varying topology to prevent segfaults.
- Adding velocity, acceleration and angular velocity attributes to the Cycles mesh

**Comments**
Added frames per second to camera [as in the volume motion blur PR](https://github.com/tangent-opensource/hdcycles/pull/75), there should not be any conflicts.

See [Cycles PR](https://github.com/tangent-opensource/cycles/pull/7)

Without motion blur
fluid_no_blur_0
With motion blur
fluid_with_blur

### DIFF
--- a/plugin/hdCycles/mesh.cpp
+++ b/plugin/hdCycles/mesh.cpp
@@ -299,8 +299,6 @@ HdCyclesMesh::_AddVelocities(VtVec3fArray& velocities,
 
         for (size_t i = 0; i < velocities.size(); ++i) {
             V[i] = vec3f_to_float3(velocities[i]);
-            //printf("Adding velocity %.3f %.3f %.3f - %.3f %.3f %.3f\n", V[i].x, V[i].y, V[i].z,
-            //velocities[i][0], velocities[i][1], velocities[i][2]);
         }
     } else {
         TF_WARN("Velocity requries per-vertex interpolation");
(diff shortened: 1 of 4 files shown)


skwerner (approved):

dedoardo:
Removed the angular velocities to reflect the change in the [Cycles side](https://github.com/tangent-opensource/cycles/pull/7).

plugin/hdCycles/mesh.cpp
```diff
@@ -272,57 +277,64 @@ HdCyclesMesh::_AddVelocities(VtVec3fArray& velocities,
...
-            const int vCount = m_faceVertexCounts[i];
+        ccl::float3* V = attr_V->data_float3();
 
-            for (int j = 1; j < vCount - 1; ++j) {
-                int v0 = *idxIt;
-                int v1 = *(idxIt + j + 0);
-                int v2 = *(idxIt + j + 1);
+        for (size_t i = 0; i < velocities.size(); ++i) {
+            V[i] = vec3f_to_float3(velocities[i]);
+            //printf("Adding velocity %.3f %.3f %.3f - %.3f %.3f %.3f\n", V[i].x, V[i].y, V[i].z,
```
bareya, line 302: You can blast these comments

bareya (approved):
looks good, maybe blast redundant comments

dedoardo:
Thanks for checking the code @bareya 

Removed comment and move the topology update to the `tag_update()` which is the more correct way of requesting the BVH to be rebuilt.

Do you think we should handle the case where the interpolation type for the velocities is constant? It's straightforward to add, but unsure how common it really is.

bareya:
Good stuff with tag update! I totally forgot about it! 

> Do you think we should handle the case where the interpolation type for the velocities is constant? It's straightforward to add, but unsure how common it really is.

Nah I think it's all good, we can make that feature when it's requested! :)

Issue #14, opened 2020-08-21
AOV support


Vochsel:
Allow arbitrary AOV's from Cycles render.

This will also likely require custom usdCycles schema.

bareya:
#84

PR #84, opened 2021-03-11
Feature/aovs

boberfly:
Refining some much-needed AOV support in HdCycles based on the branch @skwerner started. Currently the viewport is still locked to color as this is how it works inside of Blender and then you toggle which AOV to view. Houdini works differently, it wants all AOVs up-front and then you toggle which to view, and unfortunately there is no way to pass this information to Cycles. This will need a patch to Cycles to work how Houdini expects most likely.

There's also some strange bug where if you set the default display to color/Combined (beauty) and you also have DiffDir/direct-diffuse, it will always default to use DiffDir for the display, perhaps this is a Cycles bug? All other AOVs will not override the default beauty render as default...

Having said all of the above, if we render offline to an .exr or to MPlay with the HD_CYCLES_USE_TILED_RENDERING=1 environment variable, it will render correctly to what you expect, as long as your RenderVars/RenderProducts are all correctly set in the USD file.

For those of you playing at home, to make sure this env-var is working with MPlay, set the Pre-render script on the USD Render ROP node to Python and use:
`import os;os.environ['HD_CYCLES_USE_TILED_RENDERING'] = '1';`

And the currently searched for default AOVs are listed here, if you were to make your own RenderVars:
```
Blender Convention | USD Convention (RenderVar name)
Combined             color
Depth                cameraDepth
Normal               normal
IndexOB              primId
IndexMA              IndexMA
Mist                 depth
Emission             Emit
Shadow               Shadow
UV                   UV
Vector               Vector
DiffDir              DiffDir
DiffInd              DiffInd
DiffCol              DiffCol
GlossDir             GlossDir
GlossInd             GlossInd
GlossCol             GlossCol
TransDir             TransDir
TransInd             TransInd
TransCol             TransCol
VolumeDir            VolumeDir
VolumeInd            VolumeInd
```
I don't think we need to follow exact naming-convention that Blender uses here though, we could go cameCase with more descriptive names.

There is also the start of supporting Cryptomatte in this PR, however this will need to be refined still as Hydra's HDRenderBuffer can only know of 1 buffer at a time, while Cryptomattes need to have multiple ones based on depth/rank. Houdini's Karma gets around this by using a different base-class XUSD_HydraRenderBuffer here which supplies an index to an array of render-buffers:

https://github.com/sideeffects/HoudiniUsdBridge/blob/master/src/houdini/lib/H_USD/HUSD/XUSD_HydraRenderBuffer.h

So that husk will know how to write these to disk properly. If we go down this route we will need to hide it behind a pre-processor define so that there is no hard-dependency onto Houdini (it just means cryptomattes will only work in Houdini/Husk).

There's also the startings of custom AOVs, with the syntax:
```
"aovc:myCustomColor"
"aovv:myCustomValue"
```
With the AovOutput shader name being `"myCustomColor"`/`"myCustomValue"` to get detected and written to the AOV. However the AovOutput shader node will need some work so that it is successfully detected and added to the authored materials to really test that this works, for now.

### DIFF
--- a/plugin/hdCycles/material.cpp
+++ b/plugin/hdCycles/material.cpp
@@ -122,6 +122,41 @@ IsValidCyclesIdentifier(const std::string& identifier)
     return isvalid;
 }
 
+void
+ApplyPrimvarAOVs(ccl::ShaderGraph* graph)
+{
+    if (graph) {
+        auto *geo = new ccl::GeometryNode();
+        graph->add(geo);
+        // P
+        {
+            auto *aov = new ccl::OutputAOVNode();
+            aov->name = ccl::ustring("P");
+            graph->add(aov);
+            graph->connect(geo->output("Position"), aov->input("Color"));
+        }
+
+        // Pref
+        {
+            auto *attr = new ccl::AttributeNode();
+            graph->add(attr);
+            attr->attribute = ccl::ustring("Pref");
+            auto *aov = new ccl::OutputAOVNode();
+            graph->add(aov);
+            aov->name = ccl::ustring("Pref");
+            graph->connect(attr->output("Vector"), aov->input("Color"));
+        }
+
+        // Ng
+        {
+            auto *aov = new ccl::OutputAOVNode();
+            aov->name = ccl::ustring("Ngn");
+            graph->add(aov);
+            graph->connect(geo->output("True Normal"), aov->input("Color"));
+        }
+    }
+}
+
 TfTokenVector const&
 HdCyclesMaterial::GetShaderSourceTypes()
 {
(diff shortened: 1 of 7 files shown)


boberfly:
Almost ready to non-draft this one.

I've decided to keep the default blender names, I've decoupled aovName from sourceName so it really doesn't matter now.
For example, in MPlay or the EXR, it will take the aovName so it could be whatever the RenderVars define, but sourceName will need to match the token name so that Cycles understands what pass it needs to be.

boberfly:
Just to clarify with what I said earlier about using `XUSD_HydraRenderBuffer.h`, I've decided to not use this and just manually make separate RenderVars, where sourceName is the base-name eg. `CryptoObject`, and the aovName(s) are `CryptoObject00, CryptoObject01`, etc. The `cryptomatte_depth` value under Cycles->Film acts as a rank value, so a value of 6 means there will need to be 3 RenderVars that go from 00-to-02.

For custom AOVs, these can just simply be either `AOVC` (colour) or `AOVV` (value) in sourceName, and the aovName should match the exact name of the AOV shader output. The AOV shader output is currently todo so I can't test this here yet, however I've made built-in ones like P and Pref which work.

bareya:
#14

PR #88, opened 2021-03-12
Revisiting normals generation

bareya:


dedoardo (changes requested):
Looks good. Should we also recompute the tangents? They are regenerated only if the primvars are dirty.

bareya:
@dedoardo no they are being recomputed every time vertices change: 
https://github.com/tangent-opensource/hdcycles/blob/adding_opensubdiv/plugin/hdCycles/mesh.cpp#L1003

dedoardo (approved):

PR #98, opened 2021-03-18
First step with TF_DEBUG codes

dedoardo:
Issuing a possibly useful warning through its own debug code rather than a generic warning/error. This can be controlled through the environment variable `TF_DEBUG=HDCYCLES_MESH` or `TF_DEBUG=HDCYCLES_*` in the future.

In the future non fatal application-level messages should go through our custom codes and ideally the USD/Hydra warnings should disappear through better use of the API.

### DIFF
--- a/plugin/hdCycles/debug_codes.h
+++ b/plugin/hdCycles/debug_codes.h
@@ -16,7 +16,9 @@
 //  arising from the use of this software, whether in contract, tort or otherwise.
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
-#pragma once
+
+#ifndef HD_CYCLES_DEBUG_CODES_H
+#define HD_CYCLES_DEBUG_CODES_H
 
 #include <pxr/pxr.h>
 
@@ -31,3 +33,5 @@ TF_DEBUG_CODES(
 // clang-format on
 
 PXR_NAMESPACE_CLOSE_SCOPE
+
+#endif // HD_CYCLES_DEBUG_CODES_H
--- a/plugin/hdCycles/debug_codes.h
+++ b/plugin/hdCycles/debug_codes.h
@@ -1,0 +1,33 @@
+//  Copyright 2021 Tangent Animation
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied,
+//  including without limitation, as related to merchantability and fitness
+//  for a particular purpose.
+//
+//  In no event shall any copyright holder be liable for any damages of any kind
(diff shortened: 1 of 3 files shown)


plugin/hdCycles/debug_codes.h
```diff
@@ -0,0 +1,33 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied,
+//  including without limitation, as related to merchantability and fitness
+//  for a particular purpose.
+//
+//  In no event shall any copyright holder be liable for any damages of any kind
+//  arising from the use of this software, whether in contract, tort or otherwise.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+#pragma once
```
bareya, line 19: Unfortunately, `pragma once` is not part of the standard. It is supported by `gcc`, but I think we should stick to the standard `#ifndef, #define` as we do everywhere else in the project.

plugin/hdCycles/mesh.cpp
```diff
@@ -533,9 +534,9 @@ HdCyclesMesh::_PopulateNormals(HdSceneDelegate* sceneDelegate, const SdfPath& id
 
     HdInterpolation interpolation = HdInterpolationCount;
     if (!GetPrimvarInterpolation(interpolation)) {
-        // TODO: Should we autogenerate normals or let Cycles generate them?
-        // Let's shoot a warning for now
-        //TF_WARN("Failed to find interpolation for normals for: %s", id.GetText());
+        m_cyclesMesh->add_vertex_normals();
+        TF_INFO(HDCYCLES_MESH)
```
bareya, line 538: Is this info printed every time? Maybe it should be only during debug? I don't think we have to notify users about this detail.

plugin/hdCycles/debug_codes.h
dedoardo: Fixed

plugin/hdCycles/mesh.cpp
dedoardo: It's printed only if you set the env variable TF_DEBUG to match HDCYCLES_*.
Is it so common to author models without any normals?

boberfly: Yeah it's common to not have normals if the mesh was intended to be a subdivision surface. The expected behaviour would be normals that get generated by the renderer, all smooth. Saves storing normals to disk if they can be auto-generated.

dedoardo: @boberfly The case of subdivision surfaces is [already handled earlier in the function](https://github.com/tangent-opensource/hdcycles/blob/fix/populate_normals_warnings/plugin/hdCycles/mesh.cpp#L502). This is for non-subdivision meshes which don't have normals.

bareya: @boberfly When mesh us subdivided, normals are taken from limit surface.

boberfly: Hmm, I guess for regular geometry the intended behaviour would be that the renderer would still generate smooth normals. I'm not sure if Cycles already does this.

dedoardo: Yes, it does it internally before committing the geometry. Now it's explicit and notifies the user.

PR #100, opened 2021-03-19
Modern cmake

bareya:
* Remove Find* cmakes, rely on `Cycles::Cycles` target that is exported in [this](https://github.com/tangent-opensource/cycles/pull/18) PR
* rez variables are optional
* minimise current cmake


bareya:
https://github.com/tangent-opensource/cycles/pull/18

skwerner:
Need a little time to look at it closer, but I like the reduction in complexity.

bareya:
@skwerner after you review it I would like to release it with cycles latest changes.

plugin/hdCycles/renderDelegate.cpp
```diff
@@ -39,7 +39,11 @@
 #include <render/film.h>
 #include <render/integrator.h>
 
+#if BOOST_NS == hboost
```
boberfly, line 42: We don't need this header at all in renderDelegate.cpp I can't see it being used.

plugin/hdCycles/api.h
```diff
@@ -20,6 +20,8 @@
 #ifndef HD_CYCLES_API_H
 #define HD_CYCLES_API_H
 
+#include <embree3/rtcore.h>
```
boberfly, line 23: This isn't too elegant, I am using this as a temporary solution as well, but I think we should think about fixing it in cycles itself.

plugin/ndrCycles/api.h
```diff
@@ -20,6 +20,8 @@
 #ifndef NDR_CYCLES_API_H
 #define NDR_CYCLES_API_H
 
+#include <embree3/rtcore.h>
```
boberfly, line 23: Same here: This isn't too elegant, I am using this as a temporary solution as well, but I think we should think about fixing it in cycles itself.

boberfly (commented):
Hey @bareya I like this modern cmake future, less is more!

plugin/hdCycles/api.h
bareya: I know it isn't, I am not very happy about it, but that is how I enforce include search path for embree. I agree we should fix it one day on cycles end.

plugin/hdCycles/renderDelegate.cpp
bareya: I didn't look into code itself, just wanted to limit usage of pre-processor defines. I will have a look into this one.

bareya:
@boberfly Thanks!

PR #105, opened 2021-03-23
Fix warnings

skwerner:
This is a cleanup of warnings that are thrown under Linux/GCC9.


plugin/hdCycles/renderDelegate.cpp
```diff
@@ -146,9 +146,9 @@ HdCyclesRenderDelegate::GetSupportedBprimTypes() const
 void
 HdCyclesRenderDelegate::_InitializeCyclesRenderSettings()
 {
-    static const HdCyclesConfig& config = HdCyclesConfig::GetInstance();
-
 #ifdef USE_USD_CYCLES_SCHEMA
+    // static const HdCyclesConfig& config = HdCyclesConfig::GetInstance();
```
bareya, line 150: Those are not going out?

plugin/hdCycles/camera.cpp
```diff
@@ -111,30 +111,37 @@ HdCyclesCamera::HdCyclesCamera(SdfPath const& id,
...
-
     , m_interocularDistance(0.065f)
     , m_convergenceDistance(30.0f * 0.065f)
     , m_usePoleMerge(false)
     , m_poleMergeAngleFrom(60.0f * M_PI_F / 180.0f)
     , m_poleMergeAngleTo(75.0f * M_PI_F / 180.0f)
+    , m_useDof(false)
+    , m_useMotionBlur(false)
+    , m_fps(24.f)
+//  , m_transformSamples
```
bareya, line 141: Do you think we should remove those too?

plugin/hdCycles/basisCurves.cpp
```diff
@@ -288,7 +288,8 @@ HdCyclesBasisCurves::_AddColors(TfToken name, VtValue value,
                                                ccl::TypeDesc::TypeColor,
                                                ccl::ATTR_ELEMENT_CORNER_BYTE);
 
-            ccl::uchar4* cdata = attr_vcol->data_uchar4();
+            (void)attr_vcol;
```
bareya, line 291: Can you explain reason for this?

Issue #102, opened 2021-03-22
Crash on resize viewport in houdini


dedoardo:
Houdini viewport sometimes crashes on resizing, this is because the display buffer is internally resized by cycles while the pixels are copied. 

Accessing the display buffer needs to follow the same synchronization pattern used internally by the Cycles session.
Fixed currently [in this branch](https://github.com/tangent-opensource/hdcycles/tree/fix/houdini_viewport), will create a PR after a bit more testing.

PR #109, opened 2021-03-23
Lock cycles session display buffer

dedoardo:
Locking the display when reading from its pixel buffer. This caused crashes when the Houdini viewport was being resized as the display buffer was being deallocated during reading #102 

This [Cycles PR ](https://github.com/tangent-opensource/cycles/pull/20)is required.

### DIFF
--- a/plugin/hdCycles/renderParam.cpp
+++ b/plugin/hdCycles/renderParam.cpp
@@ -1400,7 +1400,7 @@ HdCyclesRenderParam::_WriteRenderTile(ccl::RenderTile& rtile)
     if (!m_useTiledRendering)
         return;
 
-    std::lock_guard<std::mutex> lock(m_cyclesScene->mutex);
+    ccl::thread_scoped_lock session_lock(m_cyclesScene->mutex);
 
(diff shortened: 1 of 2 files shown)


plugin/hdCycles/renderParam.cpp
```diff
@@ -1400,6 +1400,8 @@ HdCyclesRenderParam::_WriteRenderTile(ccl::RenderTile& rtile)
     if (!m_useTiledRendering)
         return;
 
+    std::lock_guard<std::mutex> lock(m_cyclesScene->mutex);
```
bareya, line 1403: What do you think about using aliases that are defined `util_thread.h`?

bareya (approved):

plugin/hdCycles/renderParam.cpp
skwerner: I would also prefer the alias, this is consistent with the usage of the same lock inside Cycles.

dedoardo: 👍

PR #115, opened 2021-03-25
Add NanoVDB package to HdCycles cmake build

sergeneren:
This PR adds nanovdb package (and "WITH_NANOVDB" preprocessor definition) to hdcycles build system so cycles openVDBLoader nanogrid is initialized correctly. Fixes internal tickets PT-734 and PT-836

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,6 +178,14 @@ if(${OpenVDB_FOUND})
   message("[HDCYCLES] Configured with OpenVDB")
 endif()
 
+find_package(NanoVDB)
+if(${NanoVDB_FOUND})
+  set(WITH_NANOVDB ON)
+  add_definitions(-DWITH_NANOVDB)
+  message("[HDCYCLES] Configured with NanoVDB")
+endif()
+
+
 find_package(Embree)
 if(${Embree_FOUND})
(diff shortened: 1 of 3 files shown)


bareya:
I am sorry @sergeneren, but I have to object :( There is a lot of work you did here and unfortunately it is going to be completely scratched by #100. 

To give a bigger picture, while we export targets during Cycles install, all of the defines(PUBLIC and INTERFACE) are going to end up in exported targets cmake file. Look in to cycles install directory `lib/cmake`. There is a cmake file there with `Targets` in the name. 

Also look here https://github.com/tangent-opensource/cycles/pull/18. In the exported configuration form Cycles build we store includes to nanovdb here https://github.com/tangent-opensource/cycles/blob/cmake_exported_libraries/src/cmake/CyclesConfig.cmake.in#L51. This way HdCycles becomes a build slave of Cycles build. If you pass `-WITH_NANOVDB` flag during Cycles build, it will be automatically picked up by HdCycles.

I am sorry for giving you bad news, but in this context Cycles and HdCycles modern cmake PRs solve your problems.

sergeneren:
@bareya I'm actually glad to hear that there is no need for added complexity. I think we can just close this when #100 gets merged. Thank you.

PR #117, opened 2021-03-25
Fix hdCycles crash on updating volumes PT-835

sergeneren:
Added a function to update the openvdb grid of the "HdCyclesVolumeLoader", which is cleared by the cycles "device_load_image" function. This grid is needed for constructing a mesh and was causing a crash when volumes are updated. 

Fixes PT-835

Testing procedure:

* Download the test file
* Render in viewport with cycles
* Disable transform and render a second time, HdCycles shouldn't crash anymore.

[test file](https://github.com/tangent-opensource/hdcycles/files/6207197/volume_update.zip)

### DIFF
--- a/plugin/hdCycles/openvdb_asset.h
+++ b/plugin/hdCycles/openvdb_asset.h
@@ -41,28 +41,25 @@ PXR_NAMESPACE_OPEN_SCOPE
 class HdCyclesVolumeLoader : public ccl::VDBImageLoader {
 public:
     HdCyclesVolumeLoader(const char* filepath, const char* grid_name)
-        : ccl::VDBImageLoader(grid_name), m_file_path(filepath)
+        : ccl::VDBImageLoader(grid_name)
+        , m_file_path(filepath)
     {
-        openvdb::io::File file(filepath);
-
-        try {
-            file.setCopyMaxBytes(0);
-            file.open();
-            this->grid = file.readGrid(grid_name);
-        } catch (const openvdb::IoError& e) {
-            std::cout << "LOAD ERROR\n";
-        }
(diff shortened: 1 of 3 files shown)


plugin/hdCycles/openvdb_asset.h
```diff
@@ -53,6 +53,21 @@ class HdCyclesVolumeLoader : public ccl::VDBImageLoader {
...
     }
+
+    void UpdateGrid(){
+        openvdb::io::File file(m_file_path);
+
+        try {
+            file.setCopyMaxBytes(0);
+            file.open();
+            this->grid = file.readGrid(grid_name);
+        } catch (const openvdb::IoError& e) {
```
bareya, line 64: I would improve here:
*  We could use `TF_ ` library to report those type of errors. 
* The verbosity of the message also could be improved. Filename should be included in the error message.
* Can UpdateGrid return any other exception that `IoError`? If so that should be handled as well.

skwerner (changes requested):
Looks good to me. If you add the logging/exception handling changes that Piotr mentioned, we can merge it.

sergeneren:
Thanks for the review. 

Today I discovered that this pr doesn't cover the event when multiple volumes in a scene have different dirty states. If one of the volumes have a clean state the grid won't be updated and hdCycles will still crash. I am trying to come up with another solution.

sergeneren:
I've added an override to the VDBImageLoader class' cleanup function to retain the openvdb grid. This way we don't lose it during the mesh build time. I've tested it with a couple vdb files over a span of an hour and haven't seen a crash. It should be in good position now.

bareya (approved):

PR #124, opened 2021-03-31
create resource registry to manage shader bindings

bareya:


boberfly:
This relates (shader reference counting coming in):
https://developer.blender.org/D10965

bareya:
This can be revisited in the future, after we move committing all attributes to resource registry.

PR #130, opened 2021-04-06
Hair texture coordinates

bareya:
@skwerner After some investigation I have observed, that we had two problems with texture coordinates for hair:

#### Texture coordinates are per curve only

* [Here](https://github.com/tangent-opensource/cycles/blob/master/src/render/attribute.cpp#L514) we can see that "size" of the `ATTR_STD_UV` attribute will be equivalent size to `ATTR_ELEMENT_CURVE`.

* [Here](https://github.com/tangent-opensource/cycles/blob/master/src/render/attribute.cpp#L210) size of the `ATTR_ELEMENT_CURVE` will result as `num_curves`.

#### Updating attributes to the device

* Shader request for STD_UV attribute made by Texture Coordinate Node will omit uv attributes that size does not match the size of `ATTR_STD_UV`. I debugged it  [here](https://github.com/tangent-opensource/cycles/blob/master/src/render/geometry.cpp#L669)

For the texture coordinates per vertex I create two attributes: `ATTR_STD_UV` and an attribute that matches vertex/varying interpolation. This allows us to access underlying `st` attribute that is per vertex using Attribute Node:

An example of check board texture assigned to hair, st coordinates pulled by Attribute Node:

Screenshot from 2021-04-05 15-40-36

### DIFF
--- a/plugin/hdCycles/basisCurves.cpp
+++ b/plugin/hdCycles/basisCurves.cpp
@@ -299,108 +299,83 @@ HdCyclesBasisCurves::_AddUVS(TfToken name, VtValue value,
 {
     ccl::ustring attribName = ccl::ustring(name.GetString());
 
-    // Add hair uvs
-    ccl::Attribute* attr_uv;
-    ccl::float2* uv;
+    // convert uniform uv attrib
+
+    auto fill_uniform_uv_attrib = [&attribName](auto& uvs, ccl::AttributeSet& attributes) {
+        ccl::Attribute* attr_std_uv = attributes.add(ccl::ATTR_STD_UV, attribName);
+        ccl::float2* std_uv_data    = attr_std_uv->data_float2();
+
+        for (size_t curve = 0; curve < uvs.size(); curve++) {
+            std_uv_data[curve][0] = uvs[curve][0];
+            std_uv_data[curve][1] = uvs[curve][1];
+        }
+    };
 
     if (interpolation == HdInterpolationUniform) {
         if (value.IsHolding<VtArray<GfVec2f>>()) {
-            VtVec2fArray uvs;
-            uvs = value.UncheckedGet<VtArray<GfVec2f>>();
+            VtVec2fArray uvs = value.UncheckedGet<VtArray<GfVec2f>>();
             if (m_cyclesHair) {
-                attr_uv = m_cyclesHair->attributes.add(ccl::ATTR_STD_UV,
-                                                       attribName);
-
-                uv = attr_uv->data_float2();
-                if (uv) {
-                    size_t i = 0;
-
-                    for (size_t curve = 0; curve < uvs.size(); curve++) {
-                        uv[i++] = vec2f_to_float2(uvs[curve]);
-                    }
-                }
+                fill_uniform_uv_attrib(uvs, m_cyclesHair->attributes);
             } else {
                 // @TODO: Unhandled support for deprecated curve mesh geo
-                attr_uv = m_cyclesMesh->attributes.add(ccl::ATTR_STD_UV,
-                                                       attribName);
-
-                uv = attr_uv->data_float2();
+                ccl::Attribute* attr_std_uv = m_cyclesMesh->attributes.add(ccl::ATTR_STD_UV, attribName);
             }
         } else if (value.IsHolding<VtArray<GfVec3f>>()) {
-            VtVec3fArray uvs;
-            uvs = value.UncheckedGet<VtArray<GfVec3f>>();
+            VtVec3fArray uvs = value.UncheckedGet<VtArray<GfVec3f>>();
             if (m_cyclesHair) {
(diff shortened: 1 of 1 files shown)


boberfly:
Thank you for looking into this @bareya 

I think we should somehow treat this as an enhancement much like @dedoardo 's face-varying normals patch here:
https://github.com/tangent-opensource/cycles/blob/rez_cycles-1.13.0-ta.1.9.3/src/render/attribute.cpp#L288

Perhaps we can make an ATTR_STD_CURVE_UV like this and name it "curve_uv" but treat it like "uv" internally, much like what the face-var normals patch does by naming it "Nc" ?

That way in this code:
https://github.com/tangent-opensource/cycles/blob/rez_cycles-1.13.0-ta.1.9.3/src/render/attribute.cpp#L528
It picks ATTR_ELEMENT_CURVE_KEY

And whatever the kernel code needs to do to pick the right co-ordinate at render-time eg. pick ATTR_STD_UV by default so Blender doesn't break, or prefer picking ATTR_STD_CURVE_UV if it exists.

Thoughts @skwerner @dedoardo ?

bareya:
@boberfly Unfortunately name is ignored when attribute request contains `ATTR_STD_*` that is not `ATTR_STD_NONE` ([how attribute is requested](https://github.com/tangent-opensource/cycles/blob/master/src/render/geometry.cpp#L669), [how attribute is found](https://github.com/tangent-opensource/cycles/blob/rez_cycles-1.13.0-ta.1.9.3/src/render/attribute.cpp#L587)).

Your suggestion would require modification of `AttributeRequest` created by `Texture Coordinate Node`. I think that would break few things. Currently, the request made by aforementioned node contains `ATTR_STD_UV` only, and it does not carry any name: 

```
(gdb) print req.name
$2 = {static npos = 18446744073709551615, m_chars = 0x0, static empty_std_string = {static npos = 18446744073709551615, _M_dataplus = {<std::allocator<char>> = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>}, _M_p = 0x7fac053c2558 <std::string::_Rep::_S_empty_rep_storage+24> ""}}}
(gdb) print req.std
$3 = ccl::ATTR_STD_UV
```

dedoardo:
Hi, sorry for chiming in just now.

@bareya I think you can get around the need to modify `AttributeRequest` by changing `update_attribute_element_size` to check if the geometry is a curve and if the attribute is uv and upload the @boberfly's hypotetical `ATTR_STD_CURVE_UV` rather than `ATTR_STD_UV`. You also need to change the reading side of it to switch the indexing access depending on what type of uvs they are.
It might not look like the most elegant solution, but it is not very intrusive, is memory friendly and doesn't break default behavior.

The Attribute nodes look like a fine solution, though I think it results in two buffers being uploaded to the device and I think I am missing the point of why the `ATTR_STD_UV` is needed? Shouldn't it be some kind of redirection buffer?

dedoardo:
I am not sure if more than one location needs to be changed on the reading side of `ATTR_STD_UV`. I think that changing the content/size of `ATTR_STD_UV` should make no difference to `TextureCoordinateNode`. The compilation code only assigns that uv buffer to the out node.
Whoever takes that out as input needs to be aware of it, but it's likely to be a single piece of code responsible for this indexing. If so (hopefully), only that function would have to be changed (Similar to the normals patch).

bareya:
Thanks @dedoardo for the input, I really appreciate it!

> The Attribute nodes look like a fine solution, though I think it results in two buffers being uploaded to the device and I think I am missing the point of why the ATTR_STD_UV is needed? Shouldn't it be some kind of redirection buffer?

I don't think there would be two buffers uploaded to the device. `TextureCoordinateNode` creates request for `ATTR_STD_UV`. If we replace `TextureCoordinateNode` with `AttributeNode`, then there is no request for `ATTR_STD_UV` and only attribute requested by `AttributeNode` is uploaded.

dedoardo:
@bareya Ah ok, makes sense if the nodes are replaced, thought they were concatenated somehow :)
Are you [filling the `ATTR_STD_UV`](https://github.com/tangent-opensource/hdcycles/pull/130/files#diff-96a4c6efdfd8a60b05b6a3d7cad27d2d914d25e7c99fdbd1b54ddc25895244a4R338) buffer for compatibility?

bareya:
> Are you filling the ATTR_STD_UV buffer for compatibility?

@dedoardo yes that is correct.

bareya:
Hey @dedoardo where this part:

> You also need to change the reading side of it to switch the indexing access depending on what type of uvs they are.

would it be?

dedoardo:
I think somewhere in the [`TextureMapping` node](https://github.com/tangent-opensource/cycles/blob/master/src/render/nodes.cpp#L45), which other texture nodes (like checker) derive from.
Need to dig deeper to find the actual loc.

PR #132, opened 2021-04-08
Default color AOV fix

boberfly:
When there is no AOV specified, Houdini uses the token "C.*" as default which Cycles doesn't understand fully as it doesn't do LPEs yet, so this is a workaround to use HdAovTokens->color instead which is just the string "color".

### DIFF
--- a/plugin/hdCycles/renderParam.cpp
+++ b/plugin/hdCycles/renderParam.cpp
@@ -109,11 +109,20 @@ std::array<HdCyclesAov, 3> CryptomatteAovs = {{
     { "CryptoAsset", ccl::PASS_CRYPTOMATTE, HdCyclesAovTokens->CryptoAsset, HdFormatFloat32Vec4, true },
 }};
 
+// Workaround for Houdini's default color buffer naming convention (not using HdAovTokens->color)
+const TfToken defaultHoudiniColor = TfToken("C.*");
+
 TfToken GetSourceName(const HdRenderPassAovBinding &aov) {
     const auto &it = aov.aovSettings.find(UsdRenderTokens->sourceName);
     if (it != aov.aovSettings.end()) {
         if (it->second.IsHolding<std::string>()) {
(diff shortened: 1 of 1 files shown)


plugin/hdCycles/renderParam.cpp
```diff
@@ -2170,6 +2179,7 @@ HdCyclesRenderParam::SetAovBindings(HdRenderPassAovBindingVector const& a_aovs)
         TfToken sourceName = GetSourceName(aov);
 
         for (HdCyclesAov& cyclesAov : DefaultAovs) {
+            std::cout << "sourcename: " << sourceName << '\n';
```
bareya, line 2182: Is this logging intentional?

boberfly: Nope, cheers for the keen eye spot!

bareya (approved):

plugin/hdCycles/renderParam.cpp
```diff
@@ -109,11 +109,20 @@ std::array<HdCyclesAov, 3> CryptomatteAovs = {{
     { "CryptoAsset", ccl::PASS_CRYPTOMATTE, HdCyclesAovTokens->CryptoAsset, HdFormatFloat32Vec4, true },
 }};
 
+// Workaround for Houdini's default color buffer naming convention (not using HdAovTokens->color)
+const TfToken defaultHoudiniColor = TfToken("C.*");
```
bareya, line 113: shouldn't this variable be visible only from this translation unit?

boberfly: Not sure what you mean, it's inside an anonymous namespace?

bareya: yes, static or anon.

boberfly: https://github.com/tangent-opensource/hdcycles/blob/main/plugin/hdCycles/renderParam.cpp#L56

it's in one...

PR #134, opened 2021-04-12
Don't apply AA samples when it is 0.

boberfly:
This is dependent upon https://github.com/tangent-opensource/usdcycles/pull/3

Hdcycles will use the default samples option if aa_samples is set to 0 like the previous behaviour. If aa_samples is set higher than 0, it'll be used instead in branched-path mode, which is the expected behaviour.

The progress of 400% reported is another issue which might need to be looked at closer to see what's happening there.


bareya:
@boberfly To keep things in sync I think this PR requires package.py updated. First I would bump a version here tangent-opensource/usdcycles#3. Then in this PR, so correct version of usdCycles is being pulled during the build.

PR #135, opened 2021-04-12
Arbitrary attribute for curves

bareya:
* Adding `AttributeSource` - `HdBufferSource` interface to commit resources to the scene.
* Adding support for curves.
* Adding tests.


bareya:
This PR goes along with https://github.com/tangent-opensource/cycles/pull/27

PR #136, opened 2021-04-12
Bugfix/fix warnings

skwerner:
More warnings fixes


bareya:
No ndr api fix? :)

bareya:
Also I think formatting is busted.

plugin/hdCycles/api.h
```diff
@@ -22,28 +22,4 @@
...
-#    else
-#        define HDCYCLES_API ARCH_IMPORT
-#        define HDCYCLES_API_TEMPLATE_CLASS(...)                               \
-            ARCH_IMPORT_TEMPLATE(class, __VA_ARGS__)
-#        define HDCYCLES_API_TEMPLATE_STRUCT(...)                              \
-            ARCH_IMPORT_TEMPLATE(struct, __VA_ARGS__)
-#    endif
-#    define HDCYCLES_LOCAL ARCH_HIDDEN
-#endif
-
```
boberfly, line 48: Perhaps we should keep this here to be consistent with USD-isms if we do want to export stuff. I think other delegates leave this in right?

bareya: Hey @boberfly unfortunately I must disagree. I am very happy to see this PR. I wanted to get rid of this for a very long time. MSVC linker just goes nuts about inconsistencies between what is exported on interface and implementation level.

Just to clarify, we are writing an extension to `Hd` library, we are not creating new one. Necessary exports are defined on the interface level, we don't need to re-define them.

skwerner: HdPrman is using those export macros, but I don't see any reason why it would need to.

skwerner:
I took the liberty of applying `clang-format` too.😈

There are still lots of `-Wconversion` warnings that gcc throws, besides that it's warnings-free on Linux now for me.

PR #137, opened 2021-04-12
Remove lock from tile rendering callback

boberfly:
This lock was causing a deadlock on Linux and preventing tiled rendering from working, found this while debugging the aa_samples issue.

### DIFF
--- a/plugin/hdCycles/renderParam.cpp
+++ b/plugin/hdCycles/renderParam.cpp
@@ -1456,8 +1456,6 @@ HdCyclesRenderParam::_WriteRenderTile(ccl::RenderTile& rtile)
     if (!m_useTiledRendering)
         return;
 
-    ccl::thread_scoped_lock session_lock(m_cyclesScene->mutex);
-
     const int w = rtile.w;
     const int h = rtile.h;
 


bareya:
Is scene mutex locked during the rendering? I am just worried about this little fellow:

https://github.com/tangent-opensource/hdcycles/blob/2ea6f7cf3a8cac010522f5db17d73c51a7bdaae1/plugin/hdCycles/renderParam.cpp#L1476

dedoardo:
A couple of questions: 
- Is the `ccl::Tile::RenderBuffer` shared between tiles and if so, locked? At first glance it looks shared, need to dig into the code though.
- Is there any more context to the deadlock? Is it because the caller is locking the scene already or maybe caused by synchronization with another lock?

In general, I would also like to see this lock removed.

boberfly:
This code derives from:
https://developer.blender.org/diffusion/C/browse/master/src/blender/blender_session.cpp$331
And:
https://github.com/boberfly/GafferCycles/blob/master/src/GafferCycles/IECoreCyclesPreview/Renderer.cpp#L722

And is run on multiple threads in-flight and only should be writing to their own region of the image so race conditions shouldn't happen. If they do eg. the blitting code, then it should be fixed there. A scene reset should happen if exposure/render-buffers change.

I didn't debug it extensively to know the exact details, but it absolutely prevents tiled rendering from working at all on my Linux workstation & never progresses (stuck at 1% cpu usage until I kill the process).

bareya:
It's hard to rely on `GafferCycles` implementation because synchronisation mechanism in Hydra is completely different. As far as I remember fast timeline scrubbing was causing data races.

PR #141, opened 2021-04-13
Use tbb to set number of threads

bareya:
Adding functionality to respect `-j`/`--threads` from Husk. This is a temporary solution. When SideFx exposes Husk's TfTokens they can be taken from a header distributed with Houdini.

### DIFF
--- a/plugin/hdCycles/config.cpp
+++ b/plugin/hdCycles/config.cpp
@@ -120,7 +120,6 @@ HdCyclesConfig::HdCyclesConfig()
 
     max_samples = HdCyclesEnvValue<int>("HD_CYCLES_MAX_SAMPLES", 512);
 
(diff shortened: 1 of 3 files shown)


bareya:
https://github.com/tangent-opensource/hdcycles/issues/118

bareya:
@skwerner It seems like on Cycles `TaskScheduler`'s end `tbb` is used, but not for everything. It is some sort of hybrid of `tbb` and `numa`. 

Husk sets `tbb::tas_arena`'s number of threads. The number of threads set by Husk can be intercepted by calling `tbb::this_task_arena::max_concurrency()`. For now setting `SessionParams::threads` solves the problem, but maybe we should consider solve this problem on Cycles end? For instance [here](https://github.com/tangent-opensource/cycles/blob/master/src/util/util_system.cpp#L76) is the code responsible for getting number of threads.

PR #142, opened 2021-04-15
Point cloud Cycles primitive

dedoardo:
This patch implements the `HdPoints` primitive using the new Cycles point cloud. The following features are supported:
- Point styles (SPHERE, DISC, DISC_ORIENTED) with radius/normal per point
- Motion blur through vertex velocity/acceleration (geometry) or through transforms (object)
- Particle color and transparency 

Some things that could be needed:
- Arbitrary shaders. Some of the internal shading attributes (e.g. uv) are currently not set for points, need to investigate which ones would be needed.
- Time varying radius and normals
- Motion blur through time sampled vertex positions (Least priority)

This patch requires:
- [Cycles PR](https://github.com/tangent-opensource/cycles/pull/29)
- [UsdCycles PR](https://github.com/tangent-opensource/usdcycles/pull/2)
- (For houdini) [houdini_cycles PR](https://github.com/tangent-opensource/houdini_cycles/pull/1)


bareya:
Just an observation, this code will not compile without `usdCycles` schema support.  I am getting following errors:

```
/home/bareya/tangent/dev/hdcycles_for_houdini/build/hdcycles-prefix/src/hdcycles/plugin/hdCycles/points.cpp:133:61: error: ‘usdCyclesTokens’ was not declared in this scope
                                                             usdCyclesTokens->primvarsCyclesObjectMblur,

```

Maybe we should merge following PR first?
https://github.com/tangent-opensource/hdcycles/pull/122

dedoardo:
Good catch, forgot to wrap `_ReadObjectFlags` in #ifdefs. I never compile without `usdCycles` :)  

Waiting for #122 sounds good to me.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+    // -------------------------------------
+    // -- Resolve Drawstyles
+#ifdef USE_USD_CYCLES_SCHEMA
+    if (HdChangeTracker::IsPrimvarDirty(*dirtyBits, id, usdCyclesTokens->cyclesObjectPoint_style)) {
+        VtValue point_style_ = GetPrimvar(sceneDelegate, usdCyclesTokens->cyclesObjectPoint_style);
+        if (point_style_.IsEmpty()) {
+            TF_WARN("Point style primvar exists, but is empty for point cloud %s", id.GetText());
+        } else {
+            TfToken point_style = point_style_.Cast<TfToken>().UncheckedGet<TfToken>();
+            if (point_style == usdCyclesTokens->sphere) {
```
bareya, line 592: This seems like a perfect opportunity for appending new dirty bits. A good example would be normals in `HdStMesh`: 

```
    enum DirtyBits : HdDirtyBits {
        DirtySmoothNormals  = HdChangeTracker::CustomBitsBegin,
        DirtyFlatNormals    = (DirtySmoothNormals << 1),
        DirtyIndices        = (DirtyFlatNormals   << 1),
        DirtyHullIndices    = (DirtyIndices       << 1),
        DirtyPointsIndices  = (DirtyHullIndices   << 1)
    };
```

`PropagateDirtyBits` sounds like a perfect place to implement behaviour for new dirty bits.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+    _CheckIntegrity(param);
+
+    _UpdateObject(scene, param, dirtyBits, needsRebuildBVH);
+    *dirtyBits = HdChangeTracker::Clean;
+}
+
+HdDirtyBits
+HdCyclesPoints::GetInitialDirtyBitsMask() const
+{
+    return HdChangeTracker::DirtyPoints | HdChangeTracker::DirtyTransform | HdChangeTracker::DirtyVisibility
```
bareya, line 705: Missing `DirtyPrimvar`? if we decided to add support for arbitrary attribute

plugin/hdCycles/renderParam.cpp
```diff
@@ -1737,6 +1741,23 @@ HdCyclesRenderParam::AddMesh(ccl::Mesh* a_mesh)
     Interrupt();
 }
 
+void
+HdCyclesRenderParam::AddPointCloud(ccl::PointCloud* a_pc)
```
bareya, line 1745: `ccl::PointCloud` is a polymorphic type, do we really need this separation for curve, mesh, and point cloud? They are all sub-types of `ccl::Geometry`.

plugin/hdCycles/renderParam.cpp
```diff
@@ -1737,6 +1741,23 @@ HdCyclesRenderParam::AddMesh(ccl::Mesh* a_mesh)
...
+{
+    lock_guard lock { m_cyclesScene->mutex };
+
+    if (!m_cyclesScene) {
+        TF_WARN("Couldn't add geometry to scene. Scene is null.");
+        return;
+    }
+
+    m_pointCloudUpdated = true;
```
bareya, line 1754: I am not sure what does this do. Where is it used? `m_geometryUpdated` isn't enough?

plugin/hdCycles/renderParam.cpp
```diff
@@ -1838,6 +1859,28 @@ HdCyclesRenderParam::RemoveMesh(ccl::Mesh* a_mesh)
         Interrupt();
 }
 
+void
+HdCyclesRenderParam::RemovePointCloud(ccl::PointCloud* a_pc)
```
bareya, line 1863: If we get rid of `m_pointCloudUpdated` can we remove this function and make two generic functions: `AddGeometry` and `RemoveGeometry`?

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+        return;
+    }
 
-                pointObject->random_id = i;
-                pointObject->name      = ccl::ustring::format("%s@%08x", pointObject->name, pointObject->random_id);
-                m_cyclesObjects.push_back(pointObject);
-                param->AddObject(pointObject);
-            }
-        }
+    if (!value_.IsHolding<VtFloatArray>()) {
```
bareya, line 265: `VtValue` offers `CanCast` and `Cast` methods. Appropriate "Casters" are registered in `HdCyclesAttributeSource` implementation.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
-                    if (i < m_cyclesObjects.size()) {
-                        float w                 = widths[i];
-                        m_cyclesObjects[i]->tfm = m_cyclesObjects[i]->tfm * ccl::transform_scale(w, w, w);
-                    }
-                }
-            }
+    ccl::float4* C = attr_C->data_float4();
+    if (interpolation == HdInterpolationConstant) {
+        assert(value.size() == 1);
+        const ccl::float3 v0 = vec3f_to_float3(value[0]);
```
bareya, line 312: Unnecessary conversion

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+
+    ccl::Attribute* N_attr = m_cyclesPointCloud->attributes.find(ccl::ATTR_STD_VERTEX_NORMAL);
+    if (!N_attr) {
+        N_attr = m_cyclesPointCloud->attributes.add(ccl::ATTR_STD_VERTEX_NORMAL);
+    }
+    ccl::float3* N = N_attr->data_float3();
+
+    if (interpolation == HdInterpolationConstant) {
+        assert(value.size() == 1);
+        const ccl::float3 N0 = vec3f_to_float3(value[0]);
```
bareya, line 419: Unnecessary conversion.

bareya (changes requested):
Great stuff Edoardo! I tested your PR and it does render with small code alterations such as `TypeFloat4` :)

Unfortunately, I experienced some problems that point cloud does not appear in Houdini when I click `Restart Render`. First time it does render. Second time it disappears. 

Also, there is a bit of code duplication. If you wish investing time in removing the duplication that would be awesome. 

During implementing support for arbitrary attributes for curves I implemented very generic interface for delayed Attribute creation. It could be easily extended to support `ATTR_STD` creation.

dedoardo:
@bareya Yes, I wasn't finished with the changes. Changed the target of the PR.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+    }
 
-        needs_update = true;
+    // If the points have been reset, the attributes are expected to also have been cleared
+    ccl::AttributeSet* attributes = &m_cyclesPointCloud->attributes;
+    ccl::ustring attrib_name("displayColor");
+    ccl::Attribute* attr_C = attributes->find(attrib_name);
+    bool reset_opacity     = false;
+    if (!attr_C) {
+        attr_C        = attributes->add(attrib_name, ccl::TypeDesc::TypeFloat4, ccl::ATTR_ELEMENT_VERTEX);
```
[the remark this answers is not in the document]
dedoardo: Thanks, changed.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+HdCyclesPoints::_InitializeNewCyclesPointCloud()
 {
-    HdCyclesRenderParam* param = static_cast<HdCyclesRenderParam*>(renderParam);
-
-    const SdfPath& id = GetId();
-
-    HdCyclesPDPIMap pdpi;
+    m_cyclesPointCloud = new ccl::PointCloud();
+    assert(m_cyclesPointCloud);
+    m_cyclesPointCloud->point_style = (ccl::PointCloudPointStyle)m_pointStyle;
```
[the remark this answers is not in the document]
dedoardo: Changed to local variable

dedoardo:
Fixes:
- Fixed the issue with Houdini not showing the render(uninitialized var)
- Made the point style variable local
- Use the usd casting stuff
- Add ATTR_STD_GENERATED
- Changed the type to RGBA

Questions:
- If I understand correctly, the attribute source can only be used with attributes with no semantics (custom I guess). I can't find an example of it being used in the code outside the tests, I tried forwarding all the primvars through it but get crashes (reasonably). How do I filter which attributes are custom? 
I was hoping I could use it for any attribute... How does this reduce code duplication?
- Can add dirty bits, but see comment.

plugin/hdCycles/points.cpp
dedoardo: Tried to follow the usage in Mesh, let me know if it's ok.

plugin/hdCycles/points.cpp
```diff
@@ -94,269 +97,669 @@ HdCyclesPoints::Finalize(HdRenderParam* renderParam)
...
+        std::make_pair(HdInterpolationVertex, HdPrimvarDescriptorVector {}),
+        std::make_pair(HdInterpolationVarying, HdPrimvarDescriptorVector {}),
+        std::make_pair(HdInterpolationFaceVarying, HdPrimvarDescriptorVector {}),
+    };
+
+    for (auto& info : primvars_desc) {
+        info.second = sceneDelegate->GetPrimvarDescriptors(id, info.first);
+    }
+
+    for (auto& interpolation_description : primvars_desc) {
```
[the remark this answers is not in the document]
dedoardo: How do I know which attributes are to be forwarded to Cycles?

plugin/hdCycles/points.cpp
dedoardo: The point style are propagates the dirty points, what is the point of defining this new enum?

plugin/hdCycles/points.cpp
dedoardo: Added

dedoardo:
Actually... no crashes with the attribute source. Is this the correct usage?

plugin/hdCycles/points.cpp
```diff
@@ -730,6 +730,10 @@ HdCyclesPoints::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam,
                 _PopulateVelocities(sceneDelegate, id, interpolation, value);
             } else if (description.name == HdTokens->accelerations) {
                 _PopulateAccelerations(sceneDelegate, id, interpolation, value);
+            } else {
+                HdCyclesPointCloudAttributeSource source {
+                    description.name, HdPrimvarRoleTokens->none, { value }, m_cyclesPointCloud, interpolation
+                };
```
dedoardo, line 736: ahh, forgot to `Resolve`...

bareya: You don't need to call `Resolve`. You need to grab an `HdInstance` and create shared pointer of `HdCyclesAttributeSource`. There are aliases for it. Then append that source to HdInstance. The `HdCyclesResourceRegistry` will handle committing resources for you in scene-thread safe manner. Also, it will check if `HdCyclesAttributeSource` is a valid source, if not, it will not commit that resource.

dedoardo:
Thanks for the explanation @bareya, it all makes more sense now :) I also found an example in the hair geometry, should have looked more carefully.
Have all the comments been addressed ? 

Some comments/questions:
- Some files coming out of Houdini can be vectors of __int64 (sourceprim for example). This right now triggers an error as the type is not in the hydra enumerator. I need to check the cycles side, but I think we could look to add support for it in the future. 
- Should we check the shader network of the object to see if it needs a specific attribute before adding it to the attributes?
- One thing which is pending are the points dpi. I don't really have a use case for it though. thus added a warning in the code, what do you think?

bareya:
Hey @dedoardo looks good and I can't wait to test it! 

I can answer some questions:

> Some files coming out of Houdini can be vectors of __int64 (sourceprim for example). This right now triggers an error as the type is not in the hydra enumerator. I need to check the cycles side, but I think we could look to add support for it in the future.

`VtValue` offers "casting" interface. It's quite easy to register new casting functions I move them [here](https://github.com/tangent-opensource/hdcycles/blob/main/plugin/hdCycles/attributeSource.cpp#L577). It seemed like more appropriate place to keep in `attributeSource.cpp`. For convenience I added helper functions to register new `VtArray<>` conversions. It should be as easy as:

```c++
TryRegisterValArrayCast<float, long long int>();
```

> Should we check the shader network of the object to see if it needs a specific attribute before adding it to the attributes?

I thought about it, but I never implemented that check. I very very vaguely remember from our standups that we had a conversation about need for that check, and I think you mentioned that the attribute need requirement check has no effect until very moment of committing to a device. To be honest I didn't have that much time to go through Cycles' code to actually confirm it, but I will rely on your expertise. Feel free to educate me. Also, If you think we need it to reduce memory consumption, please do it :)

PR #150, opened 2021-04-22
Use cycles locking and removing lock/unlock mechanism 

bareya:


plugin/hdCycles/basisCurves.cpp
```diff
@@ -553,9 +553,7 @@ HdCyclesBasisCurves::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderP
 
     if (generate_new_curve) {
         if (m_cyclesGeometry) {
-            scene_lock.unlock();
             param->RemoveGeometry(m_cyclesHair);
```
dedoardo, line 556: Shouldn't this be `RemoveGeometrySafe` if there is a lock around Sync?

dedoardo:
Looks ok, the change is limited to constructors and destructors where execution should be sequential. 
Pointed out a possible mistake in the curves sync.

plugin/hdCycles/basisCurves.cpp
bareya: No because at the beginning of Sync function we have `thread_scoped_lock`.

PR #151, opened 2021-04-22
Always set default values at the start of a sync

boberfly:
Always set default values at the start of a sync, so that when a "do nothing" is set in Solaris, or a different node is picked, it'll start back with sane defaults. This addresses curves, lights, meshes and volumes.

TODO: All global values in renderParam need the defaults set on any scene refresh also, but for now we can just be explicit with our RenderSettings node (seems like no other delegate works correctly here as well, including PRMan).

And as an extra note, we may need to be vigilant on any new schema that gets added here. We can also use the cycles node API to set built-in defaults if we prefer also, but lets do that once we upgrade to the latest cycles API just in case they have a new/different way now.

### DIFF
--- a/plugin/hdCycles/basisCurves.cpp
+++ b/plugin/hdCycles/basisCurves.cpp
@@ -409,6 +409,13 @@ HdCyclesBasisCurves::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderP
     bool generate_new_curve = false;
     bool update_curve       = false;
 
+    // Defaults
+    m_visCamera = m_visDiffuse = m_visGlossy = m_visScatter = m_visShadow = m_visTransmission = true;
+    m_useMotionBlur = false;
+    m_cyclesObject->is_shadow_catcher = false;
+    m_cyclesObject->pass_id = 0;
+    m_cyclesObject->use_holdout = false;
+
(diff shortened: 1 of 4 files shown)


plugin/hdCycles/basisCurves.cpp
```diff
@@ -409,6 +409,13 @@ HdCyclesBasisCurves::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderP
     bool generate_new_curve = false;
     bool update_curve       = false;
 
+    // Defaults
+    m_visCamera = m_visDiffuse = m_visGlossy = m_visScatter = m_visShadow = m_visTransmission = true;
+    m_useMotionBlur = true;
```
bareya, line 414: Why motion blur is enabled by default?

boberfly: As we mentioned in our chat, I was only setting to true because your other patch will be defaulting it to true. I'm happy to make it false.

bareya: My patch will not be defaulting it to true on object level, but on integrator level.

boberfly: Changed to false now

PR #152, opened 2021-04-23
Distribution enum value out of range and fallback to default GGX

bareya:
Certain scenes contain materials that the index is out of range. I am not sure if the check `index < socket.size()` is the right way to go to prevent scenes from crashing.


bareya:
This needs some polishing.

PR #158, opened 2021-04-28
Changed the tile_size render setting schema type to integer 

dedoardo:
Tile size was read as GfVec2i while the schema defines it as GfVec2f. Issuing a warning for now. 
Maybe this should be automatically converted, but strict compliance with the schema is also good.

Internal ticket #6655

### DIFF
--- a/plugin/hdCycles/renderParam.cpp
+++ b/plugin/hdCycles/renderParam.cpp
@@ -463,14 +463,18 @@ HdCyclesRenderParam::_HandleSessionRenderSetting(const TfToken& key, const VtVal
     // Tiles
 
     if (key == usdCyclesTokens->cyclesTile_size) {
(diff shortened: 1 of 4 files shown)


bareya:
Hey @dedoardo, don't you think this is a mistake in our schema? Schema should conform what is defined in `SessionParams`, don't you think? I checked and `tile_size` is `int2` shouldn't we modify schema instead?

dedoardo:
@bareya Good point, I hesitated to modify the schema as I thought it would break existing publishes without automatic conversion. The compliance with cycles types is valuable though. Maybe editing the schema is a better option.

Floating points tile sizes don't really make a lot of sense do they? :)

dedoardo:
Changed the schema and warning the user if converting from GfVec2f

bareya:
@dedoardo nice one!

Issue #111, opened 2021-03-23
Houdini grid overlapping render output


dedoardo:
The world grid overlay (with transparency) in Houdini overlaps the rendering output. I would guess the depth buffer needs to accessible to Houdini to correctly compose the two images.

Would be nice to fix, though not an high priority.

![Capture](https://user-images.githubusercontent.com/2072636/112228490-c4d6c000-8c07-11eb-8830-8522185a6e1d.PNG)

boberfly:
Yeah the depth AOV needs added, and I think this will help camera movement also as it's not working right either. I'd like to move the priority up but we can get by with GL mode as a workaround...

dedoardo:
After PR #162, this has become only a Linux specific issue.

PR #162, opened 2021-05-04
Display AOVs rework

dedoardo:
This patch introduces the following changes when rendering in non-tiled mode. It requires [this Cycles PR](https://github.com/tangent-opensource/cycles/pull/42).
- Aov's render buffers are now updated once a rendering pass has finished through the new callback
- Blitting render buffers now does a single copy rather than 2 and is not tied to `RenderPass::_Execute` anymore.
- Misc improvements to `HdRenderBuffer`.
  - Memory is actually deallocated on deallocate and shrunk on resize
  - Access is now synchronized with resize / delete operations  
  - Better lifetime management with the renderparam through `BPrim::Finalize` 
- Picking and viewport overlays are also working now by slightly relaxing `GetSourceName` 

(edit: Rethinking about it, this might not be an issue at all but indicate that the time to first pixel is lower)
One **Issue** is that on extremely simple scenes you can notice that the first frame displayed is lower resolution than typically. My intuition is that it is because  blitting to the `HdRenderBuffer` happens exclusively with rendering and at the same time requires locking access to the Cycles Render Buffers. This hypothesis is (just partly) sustained by one performance metric. The new blitting code path is substantially faster than the one going through the device tasks, even if not fully optimized (the old code path only blitted one):

| Function | Time   |
| --- | --- |
| `copy_to_display_buffer` (Old) | 0.03 +- 0.001 s |
| `on_display_cb`               (New)|  0.001 +- 0.0005 s |

The tiled and adaptive code path could be further unified together, this is a first step.

Houdini grid overlay
image

Picking
image

Viewport AOVs
image


plugin/hdCycles/renderParam.cpp
```diff
@@ -2180,7 +2180,7 @@ HdCyclesRenderParam::BlitFromCyclesPass(const HdRenderPassAovBinding& aov, int w
 
             // todo: Is there a utility to convert HdFormat to string?
             if (pixels_type == ccl::RenderBuffers::ComponentType::None) {
-                TF_WARN("Unsupported component type %d for aov %s ", rb->GetFormat(), aov.aovName.GetText());
+                TF_WARN("Unsupported component type %d for aov %s ", (int)rb->GetFormat(), aov.aovName.GetText());
```
bareya, line 2183: As a small comment I would rely on C++ static_cast feature, rather than old style c-style casts. Those casts happen at runtime without any checks. I strongly recommend using static_casts.

dedoardo:
Thinking about it, the fact that we see a lower resolution frame earlier should mean that it's more responsive.

boberfly:
Doing a stress-test here @dedoardo and adding/removing a bunch of AOVs seems to work great, no crashes. It does get slower the more AOVs you load in but that's definitely to be expected, it feels fast enough with Houdini's default color+depth+normal loaded in though and we can be mindful of making custom viewport settings which don't load all AOVs at once in the default case...

My little patch tweak uses ccl::PASS_DEPTH instead of ccl::PASS_MIST and we discovered it wasn't quite what Hydra expects and has a weird default of the depth value set to 100.0f instead of 1.0f (which threw off the camera tumble code in the viewport, which must use depth for camera calculations).

Awesome work @dedoardo !

boberfly:
Just an update, the depth seems to work on windows but I am not sure why linux doesn't seem to work quite right yet...

plugin/hdCycles/renderBuffer.cpp
```diff
@@ -171,62 +195,22 @@ HdCyclesRenderBuffer::SetConverged(bool cv)
...
-    }
-}
-
 void
 HdCyclesRenderBuffer::Clear()
 {
     if (m_format == HdFormatInvalid)
         return;
 
+    m_mutex.lock();
```
skwerner, line 204: Scoped lock

plugin/hdCycles/renderBuffer.cpp
```diff
@@ -79,29 +80,43 @@ HdCyclesRenderBuffer::HdCyclesRenderBuffer(HdCyclesRenderDelegate* renderDelegat
...
     if (dimensions[2] != 1) {
         TF_WARN("Render buffer allocated with dims <%d, %d, %d> and format %s; depth must be 1!", dimensions[0],
                 dimensions[1], dimensions[2], TfEnum::GetName(format).c_str());
         return false;
     }
 
-    m_width = static_cast<unsigned int>(dimensions[0]);
-    m_height = static_cast<unsigned int>(dimensions[1]);
-    m_format = format;
+    m_mutex.lock();
```
skwerner, line 106: scoped lock

plugin/hdCycles/renderBuffer.cpp
```diff
@@ -288,13 +272,18 @@ HdCyclesRenderBuffer::BlitTile(HdFormat format, unsigned int x, unsigned int y,
 void
 HdCyclesRenderBuffer::_Deallocate()
 {
-    m_wasUpdated = true;
-    m_width = 0;
-    m_height = 0;
-    m_format = HdFormatInvalid;
-    m_buffer.resize(0);
+    m_mutex.lock();
```
skwerner, line 275: scoped lock

skwerner (changes requested):
Can you run clang-format on this patch?

dedoardo:
Thanks for reviewing @skwerner, ran clang-format and changed the scoped lock.

boberfly:
Hey I've got a screenshot of the latest patch here side by side windows vs linux, not sure why the grid doesn't get culled on the linux side (visually the AOVs are being made when I toggle depth/normal so they do exist and are working):
depth_windows

dedoardo:
After the discussion today I am merging this for now and keeping #111 alive, making it linux specific

PR #169, opened 2021-05-06
Added map_resolution schema for dome lights importance sampling

boberfly:
- Fixed lights not being square-sampled when enabled in the render settings

This PR will require houdini_cycles to be updated to show the map_resolution schema

### DIFF
--- a/plugin/hdCycles/renderParam.h
+++ b/plugin/hdCycles/renderParam.h
@@ -355,6 +355,14 @@ public:
 
     void UpdateShadersTag(ccl::vector<ccl::Shader*>& shaders);
 
+    /**
+     * @brief Return if square samples is enabled (for light samples to check against)
+     * 
+     * @return boolean
+     * 
+     */
+    bool IsSquareSamples() const { return m_useSquareSamples; }
+
 private:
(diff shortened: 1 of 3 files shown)


plugin/hdCycles/renderParam.h
```diff
@@ -355,6 +355,14 @@ class HdCyclesRenderParam : public HdRenderParam {
 
     void UpdateShadersTag(ccl::vector<ccl::Shader*>& shaders);
 
+    /**
+     * @brief Return if square samples is enabled (for light samples to check against)
+     * 
+     * @return boolean
+     * 
+     */
+    bool IsSquareSamples() const { return m_useSquareSamples; }
```
bareya, line 364: Where is this being set?

boberfly: Existing schema has a bool for it here:
https://github.com/tangent-opensource/hdBlackbird/blob/main/plugin/hdCycles/renderParam.cpp#L333

This is just a getter for light.cpp when syncing the light. Arguably we might need to flush a re-sync somehow of the lights if that global schema changes...

PR #189, opened 2021-05-15
handle invalid meshes

bareya:
Handling non manifold meshes that fail to be refined because broken topology or whatever reason

### DIFF
--- a/plugin/hdCycles/mesh.cpp
+++ b/plugin/hdCycles/mesh.cpp
@@ -684,6 +684,10 @@ HdCyclesMesh::_PopulateTopology(HdSceneDelegate* sceneDelegate, const SdfPath& i
         m_cyclesMesh->triangles[i] = refined_indices[i];
     }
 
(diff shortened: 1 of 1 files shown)


plugin/hdCycles/mesh.cpp
```diff
@@ -684,6 +684,10 @@ HdCyclesMesh::_PopulateTopology(HdSceneDelegate* sceneDelegate, const SdfPath& i
         m_cyclesMesh->triangles[i] = refined_indices[i];
     }
 
+    for (size_t i = 0; i < m_cyclesMesh->verts.size(); ++i) {
```
skwerner, line 687: ::memset() instead?

skwerner (commented):
Looks good to me.

plugin/hdCycles/mesh.cpp
bareya: Oh I didn't see that comment, yeah I should replace it with `memset`!

PR #191, opened 2021-05-17
Feature/texture cache

skwerner:
Texture cache integration. Mostly just passing through the necessary variables.


plugin/hdCycles/renderParam.cpp
```diff
@@ -598,7 +598,10 @@ HdCyclesRenderParam::_UpdateSceneFromConfig(bool a_forceInit)
 
     sceneParams->bvh_layout = ccl::BVH_LAYOUT_EMBREE;
 
-    sceneParams->persistent_data = true;
+    sceneParams->persistent_data = false;
```
skwerner, line 601: This is intentional. Going back between caching on/off with persistent data turned on can lead to crashes.

plugin/usdCycles/schema.usda
```diff
@@ -229,6 +229,105 @@ class "CyclesRendererSettingsAPI" (
...
+
+    uniform bool cycles:texture_use_custom_path = 0 (
+        customData = {
+            string apiName = "texture_use_custom_path"
+        }
+        displayGroup = "Texture Cache"
+        displayName = "Use custom cache path"
+        doc = ""
+    )
+
```
skwerner, line 330: Is there a way I can add a generic string, or even better, a file path there?

bareya: Grepping through the USD repo I found [this](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/usdRi/schema.usda#L100):

```
class RisObject "RisObject" (
    inherits = [</Shader>]
    doc = """
    \\deprecated Specialized RIS shader schemas have been deprecated in 
    favor of all shader prims being simple UsdShadeShader.
    This schema will be removed in a future release.

    Represents a ris object with connectable parameters. """ 
) {
    asset info:filePath (
        customData = { 
            string apiName = "filePath"
        }
    )
    asset info:argsPath (
        customData = {
            string apiName = "argsPath"
        }
    )
}
```

Looks like something that you could try.

bareya (approved):

plugin/usdCycles/schema.usda
boberfly: Hey @skwerner where you have = 0  you can specify a string there. Although it's a bool, would you need a different one to specify a string, or is the bool here a mistake?

Also I am not sure if uniform is what we want/need, @bareya are we able to omit uniform here as it's a single attribute? I couldn't find much info in the documentation about when we need uniform or not, like if it's an array?

bareya: It seems like `uniform`, `constant` keywords are meaningful only in context of Primvar interpolation, and they are defined by UsdGeom schema. This keyword seems inappropriate since we inherit from `APISchemaBase`.

PR #196, opened 2021-05-19
Revert "Display AOVs rework"

bareya:
Reverts tangent-opensource/hdBlackbird#162

There is severe multi-threaded performance hit on this PR, I think it's not production ready.

### DIFF
--- a/plugin/hdCycles/material.cpp
+++ b/plugin/hdCycles/material.cpp
@@ -806,7 +806,6 @@ HdCyclesMaterial::IsValid() const
     return true;
 }
 
-// clang-format off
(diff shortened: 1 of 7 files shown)


boberfly:
Some beckmark findings on a light scene (a sphere):
- Before patch (0:04)
- AOV patch (0:13)
- We blit like 4 buffers now, artificially limit to colour in the code (0:07)

Some benchmark findings on a heavy scene:
- Before patch (1:03, 75-80% per-core utilisation)
- AOV patch (1:11 65-70% per-core)
- Colour-only (1:05)

PR #200, opened 2021-05-21
instanced objects now respect their source transform.

skwerner:


bareya:
Good stuff, merging so it would make to the next release

PR #201, opened 2021-05-21
Implementing first part of denoise

boberfly:
- Simplified schema & removed legacy options from SessionParams (now in DenoiseParams), removed NLM as this will be deprecated in future Cycles releases (let me know @skwerner  if this is ok to do)
- Can render DenoiseNormal and DenoiseAlbedo
- No actual denoising is done in the renderer (yet)


skwerner:
Removing NLM is fine.

PR #202, opened 2021-05-21
Motion blur schema update and multisegment deformation motion blur

bareya:
* new schema for deformation and transformation motion blur
* by default motion blur is on
* transformation samples set to 3
* deformation samples set to 3
* individual control over motion blur, deformation, transformation samples per object

### DIFF
--- a/plugin/hdCycles/basisCurves.cpp
+++ b/plugin/hdCycles/basisCurves.cpp
@@ -98,9 +98,9 @@ HdCyclesBasisCurves::HdCyclesBasisCurves(SdfPath const& id, SdfPath const& insta
     , m_renderDelegate(a_renderDelegate)
 {
     static const HdCyclesConfig& config = HdCyclesConfig::GetInstance();
-    config.motion_blur.eval(m_useMotionBlur, true);
+    config.motion_blur.eval(m_motionBlur, true);
 
-    m_cyclesObject = _CreateObject();
+    m_cyclesObject = new ccl::Object();
 
     // when time comes to switch object management to resource registry we need to switch off reference
     auto resource_registry = dynamic_cast<HdCyclesResourceRegistry*>(m_renderDelegate->GetResourceRegistry().get());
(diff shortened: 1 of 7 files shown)


plugin/hdCycles/basisCurves.cpp
```diff
@@ -180,28 +169,39 @@ HdCyclesBasisCurves::_PopulateCurveMesh(HdRenderParam* renderParam)
...
 
+    m_cyclesHair->use_motion_blur = true;
+    m_cyclesHair->motion_steps = static_cast<unsigned int>(numSamples + ((numSamples % 2) ? 0 : 1));
+
+    attr_mP = attributes->add(ccl::ATTR_STD_MOTION_VERTEX_POSITION);
     ccl::float3* mP = attr_mP->data_float3();
-    for (size_t i = 0; i < m_pointSamples.count; ++i) {
-        if (m_pointSamples.times.data()[i] == 0.0f) {
+
+    for (size_t i = 0; i < motion_samples.count; ++i) {
```
dedoardo, line 199: Could this go through the Attribute source and be resampled like transforms?

dedoardo: Ah, saw the todo in the mesh class :+1:

boberfly:
Visually LGTM but I don't have time currently to test thorough. I am wondering though if Hydra is smart enough to not make everything deform motion blur of 3 steps if it's a static non-changing geo and we now default to it. Cheers!

bareya:
@boberfly In `Hydra` we receive samples, we don't request them from Scene Delegate. You can request 3 transformation samples, but SamplePrimvar is going to receive only 1 and then there is no motion blur.

plugin/hdCycles/mesh.cpp
```diff
@@ -615,30 +608,35 @@ HdCyclesMesh::_PopulateMotion(HdSceneDelegate* sceneDelegate, const SdfPath& id)
...
+    auto& times = motion_samples.times;
+    auto& values = motion_samples.values;
+
+    ccl::AttributeSet* attributes = &m_cyclesMesh->attributes;
+    ccl::Attribute* attr_mP = attributes->find(ccl::ATTR_STD_MOTION_VERTEX_POSITION);
+    if (attr_mP) {
+        attributes->remove(attr_mP);
+    }
 
     if (numSamples <= 1) {
```
bareya, line 625: @boberfly here we terminate deformation motion blur population, if number of samples is insufficient.

PR #207, opened 2021-05-26
Handling crop render settings (dataWindowNDC)

dedoardo:
Uses the dataWindowNDC attribute as cropping region for coreBlackbird.

It requires this [coreBlackbird PR](https://github.com/tangent-opensource/coreBlackbird/pull/58)

Picture

### DIFF
--- a/plugin/hdCycles/renderBuffer.cpp
+++ b/plugin/hdCycles/renderBuffer.cpp
@@ -214,8 +214,7 @@ HdCyclesRenderBuffer::Finalize(HdRenderParam* renderParam)
 
 void
 HdCyclesRenderBuffer::BlitTile(HdFormat format, unsigned int x, unsigned int y, int unsigned width, unsigned int height,
-                               float width_data, float height_data,
-                               int offset, int stride, uint8_t const* data)
+                               float width_data, float height_data, int offset, int stride, uint8_t const* data)
 {
     // TODO: BlitTile shouldnt be called but it is...
     if (m_width <= 0) {
@@ -230,7 +229,6 @@ HdCyclesRenderBuffer::BlitTile(HdFormat format, unsigned int x, unsigned int y, 
 
     size_t pixelSize = HdDataSizeOfFormat(format);
 
-    // printf("src tile %d %d - %d %d | %d %d\n", x, y, width, height, width_data, height_data);
     const float x_scale_dst = (float)m_width / width_data;
     const float y_scale_dst = (float)m_height / height_data;
     const unsigned int x_dst = round(x_scale_dst * x);
--- a/plugin/hdCycles/renderBuffer.cpp
+++ b/plugin/hdCycles/renderBuffer.cpp
@@ -214,6 +214,7 @@ HdCyclesRenderBuffer::Finalize(HdRenderParam* renderParam)
 
 void
 HdCyclesRenderBuffer::BlitTile(HdFormat format, unsigned int x, unsigned int y, int unsigned width, unsigned int height,
+                               float width_data, float height_data,
                                int offset, int stride, uint8_t const* data)
 {
     // TODO: BlitTile shouldnt be called but it is...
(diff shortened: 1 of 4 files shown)


plugin/hdCycles/renderParam.cpp
```diff
@@ -1208,6 +1210,24 @@ HdCyclesRenderParam::_HandleFilmRenderSetting(const TfToken& key, const VtValue&
                                                                 false);
     }
 
+    if (key == "dataWindowNDC") {
```
dedoardo, line 1213: How to access the USD tokens from Hydra?

dedoardo:
Viewport should work fine, mplay necessitates some debugging.

plugin/hdCycles/renderParam.cpp
boberfly: It's in the header `pxr/usd/usdRender/tokens.h`
`UsdRenderTokens->dataWindowNDC;`

dedoardo:
With this last commit, this patch behaves like Karma both in viewport mode and when rendering with mplay.

dedoardo:
Both tiled and non-tiled rendering now behave identically and separate the resolution of the data with that of the display.
Whenever data needs to be copied to the display, it is done with nearest point sampling. This is consistent with Karma both in the viewport and mplay.

For a `ndcDataWindow` of `(0.5, 0.1, 0.9, 0.9)`, the render pass is stretched to fit the viewport
Screenshot from 2021-05-27 11-13-35

while in mplay it preserves the aspect ratio. This is because the size of the aov is already cropped.
Screenshot from 2021-05-27 11-14-38

Are there other (or different) behavior requirements ?

boberfly:
Hey @dedoardo 
- I haven't had a chance to see how Karma does cropped windows for the viewport, but does it also stretch like that? Seems strange but not too fussed, it's the MPlay use case which is looking like it's working great. I'd imagine the first screenshot seems more useful than stretching, but if Karma does the stretch also, then that's fine.

- As for behaviour requirements, I guess you are treating overscan as a separate task to this one?

Great work, the team will be happy!

dedoardo:
@boberfly 
- It seemed strange to me too, but it's simple to verify. I do think having the cropping as in the original screenshot is much nicer (functional too?). It's something we can look into if there are specific requirements. I wonder if there are tools within the houdini viewport that let you select a rect as render region. If so, it should work out of the box (tm).
- Yes, overscan will come in a separate PR. I guess there is no particular rush to review/merge this for now.

boberfly:
Hmm if it's stable for you now and it matches Karma, I think we can merge it. I'll take a look at the code after some meetings.

plugin/hdCycles/renderParam.cpp
```diff
@@ -1261,6 +1295,19 @@ HdCyclesRenderParam::_HandleFilmRenderSetting(const TfToken& key, const VtValue&
 
     if (film_updated) {
         film->tag_update(m_cyclesScene);
+
+        // todo: Should this live in another location?
```
boberfly, line 1299: Yeah I guess we could have our own dedicated reset buffers function. I think it's ok for now, maybe @bareya might think differently but I think lets get this running on the floor.

plugin/hdCycles/renderBuffer.cpp
```diff
@@ -202,7 +202,7 @@ HdCyclesRenderBuffer::Clear()
     std::lock_guard<std::mutex> lock { m_mutex };
 
     size_t pixelSize = HdDataSizeOfFormat(m_format);
-    memset(&m_buffer[0], 0, m_buffer.size() * pixelSize);
+    memset(&m_buffer[0], 0, m_buffer.size());
```
boberfly, line 205: Nice fix, is the pixelSize line above still needed here?

boberfly (approved):
Just some notes @dedoardo - just relying on you testing this out without crashes... :)

PR #209, opened 2021-05-28
Fixed possible NULL pointer dereference

skwerner:
When encountering shader nodes that could not be converted to a
Blackbird equivalent, the code was still trying to operate on them.


bareya:
Since we use C++ `RTTI` shouldn't we use `nullptr` instead of c-style `NULL`? `NULL` is just `0`, where `nullptr` is a type of `nullptr_t`.

skwerner:
We should do the change from NULL to nullptr in a separate commit IMHO. This fix is moving code from the outside to the inside of an if{} block.

PR #210, opened 2021-05-28
Bugfix/fix warnings

skwerner:
Fixes a few warnings


cmake/FindHoudiniUsd.cmake
```diff
@@ -98,25 +98,22 @@ endforeach()
...
-find_program(USD_SCHEMA_GENERATOR
-        NAMES
-        usdGenSchema
-        PATHS
-        ${HOUDINI_ROOT}/bin
-        NO_DEFAULT_PATH
-        )
-
-# Fallback to py script, remove after 18.5.519 release and add REQUIORED to usdGenSchema find_program
+if(APPLE)
```
bareya, line 101: @skwerner can't you do exactly same thing as you did in 110 line? Instead of adding `HOUDINI_BIN` add `${HOUDINI_ROOT}/Resources/bin` to `PATHS` in `find_program`?

bareya (changes requested):

bareya (approved):

PR #211, opened 2021-05-31
Feature/dynamic volume load

sergeneren:
This PR allows hdBlackBird to import vdb grids directly from Houdini memory. 

Test: 
A file with a vdb grid imported by SOP_Import or a one created with SOP_Create should induce a crash in main but should render correctly with this pr. If the grid is a volume loaded from a vdb file, loading is done the old way.


plugin/hdCycles/houdini/houdini_vdb_loader.h
```diff
@@ -0,0 +1,52 @@
...
+#ifndef HOUDINI_VDB_LOADER_H
+#define HOUDINI_VDB_LOADER_H
+
+#include "pxr/pxr.h"
+
+#include <openvdb/openvdb.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+class HoudiniVdbLoader {
```
bareya, line 29: Does this class deserve public interface? I think it can be an anon class hidden in `open_vdbasset.cpp`, don't you think? Or is there a reason to make this public?

plugin/hdCycles/houdini/houdini_vdb_loader.h
```diff
@@ -0,0 +1,52 @@
...
+public:
+    static HoudiniVdbLoader const& Instance() {
+        static HoudiniVdbLoader instance;
+        return instance;
+    }
+
+    ~HoudiniVdbLoader();
+
+    openvdb::GridBase::ConstPtr const GetGrid(const char* filepath, const char* name) const;
```
bareya, line 38: Second `const` is redundant. `ConstPtr` is `shared_ptr<const X>`. After `shared_ptr<const X>` is declared `const` it `const` correctnes can be removed by copying the `shared_ptr`. For instance:

```c++
openvdb::GridBase::ConstPtr const my_grid1 = loader.GetGrid(file_path, name );
openvdb::GridBase::ConstPtr my_grid2 = my_grid;
```

plugin/hdCycles/houdini/houdini_vdb_loader.h
```diff
@@ -0,0 +1,52 @@
...
+
+private:
+    HoudiniVdbLoader();
+
+private:
```
bareya, line 43: Two private sections? what for?

plugin/hdCycles/CMakeLists.txt
```diff
@@ -119,6 +119,23 @@ if (MSVC)
...
+        houdini/houdini_vdb_loader.cpp
+        houdini/houdini_vdb_loader.h
+        )
+
+target_sources(hdCycles
+        PRIVATE
+        ${HOUDINI_VDB_SOURCES}
+        )
+		
+target_compile_definitions(hdCycles
```
bareya, line 133: I would stay away from defining new marcos like that in the hdCycles scope. This is Houdini specific stuff. Wouldn't be easier to modify following [line](https://github.com/tangent-opensource/hdBlackbird/blob/main/cmake/FindHoudiniUsd.cmake#L28) to:

```cmake
target_compile_definitions(UsdInterface
            INTERFACE
            USE_HBOOST
            BOOST_NS=hboost
            Houdini_FOUND=TRUE
        )
```

`Houdini_FOUND` name is taken from HoudiniCmake config(name is not accidental). Then You have one unified variable to control Houdini stuff, only being present when you link with HoudiniUsd cmake interface.

plugin/hdCycles/openvdb_asset.h
```diff
@@ -32,6 +32,9 @@
 #ifdef WITH_OPENVDB
 #    include <openvdb/openvdb.h>
 #    include <render/image_vdb.h>
+#   ifdef USE_HOUDINI_VDB_LOADER
+#       include "houdini/houdini_vdb_loader.h"
```
bareya, line 36: If this interface is private, and I think it is, this should go away. Also, if you decide this interface should be public, the include should be moved to cpp file to remove header dependency.

plugin/hdCycles/houdini/houdini_vdb_loader.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "houdini_vdb_loader.h"
+
+#include "pxr/base/arch/library.h"
+#include "pxr/base/tf/diagnostic.h"
+
+#include <GT/GT_PrimVDB.h>
+
+#ifdef WIN32
+#include <Windows.h>
+#define GETSYM(handle, name) GetProcAddress((HMODULE)handle, name)
```
bareya, line 29: Those defines are redundant. They are already defined [here](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/arch/library.h#L95) and [here](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/arch/library.cpp#L90).

plugin/hdCycles/houdini/houdini_vdb_loader.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "pxr/pxr.h"
+
+#include <openvdb/openvdb.h>
+
+PXR_NAMESPACE_OPEN_SCOPE
+
+class HoudiniVdbLoader {
+public:
+    static HoudiniVdbLoader const& Instance() {
```
bareya, line 31: Does this really have to be singleton? It could be private function initializing static variable.

plugin/hdCycles/houdini/houdini_vdb_loader.h
```diff
@@ -0,0 +1,52 @@
...
+//
+//  In no event shall any copyright holder be liable for any damages of any kind
+//  arising from the use of this software, whether in contract, tort or otherwise.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+#ifndef HOUDINI_VDB_LOADER_H
+#define HOUDINI_VDB_LOADER_H
+
+#include "pxr/pxr.h"
```
bareya, line 23: This should be in `<>` https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

plugin/hdCycles/openvdb_asset.h
```diff
@@ -79,6 +55,11 @@ class HdCyclesVolumeLoader : public ccl::VDBImageLoader {
 
 private:
     std::string m_file_path;
+#    ifdef Houdini_FOUND
```
bareya, line 58: Last little question. Is `m_sopVolLibHandle` going to change between instances of of `ccl::VdbImageLoader`? Wouldn't be better to make it a static variable in cpp file? It seems like that handle has to be initialized only once and reused for application lifetime.

sergeneren: Moved the logic to an anonymous namespace. Vdb loader function is now a static variable

bareya (approved):
Beautiful! :)

cmake/FindHoudiniUsd.cmake
```diff
@@ -29,6 +29,7 @@ target_compile_definitions(UsdInterface
     INTERFACE
     USE_HBOOST
     BOOST_NS=hboost
+    Houdini_FOUND=TRUE
```
bfloch, line 32: Is this a common convention used somewhere else?
Just unusual to end up with mixed-case definitions.
Used to USE_HOUDINI or WITH_HOUDINI kind of definitions.

bfloch (approved):
Just a minor question.
Looks good to me, thanks for the License update.

cmake/FindHoudiniUsd.cmake
sergeneren: hey @bfloch, it's coming from Houdini's cmake file HoudiniConfig.cmake

PR #212, opened 2021-06-02
Fix merge bug

dedoardo:
Last two PRs merged #210 and #208 addressed the same warning in non compatible ways.


bareya (approved):
makes sense!

PR #213, opened 2021-06-02
Mesh motion normals

dedoardo:
This patch populates custom motion normals for meshes. It requires this [Cycles PR](https://github.com/tangent-opensource/coreBlackbird/pull/60)

Testing it on assets with motion blur/subdivision and seems to behave as expected. Will do some further tests before merging, but it's ready to be reviewed.

@bareya Does the current solution suffice for now? Might not be super elegant but at least avoids duplicating too much code. We will eventually have to revisit the timesampled primvars

### DIFF
--- a/plugin/hdCycles/mesh.cpp
+++ b/plugin/hdCycles/mesh.cpp
@@ -712,18 +712,7 @@ HdCyclesMesh::_PopulateMotionAttributeVec3f(HdSceneDelegate* sceneDelegate, cons
         if (times[i] == 0.0f)  // todo: more flexible check?
             continue;
 
-        VtValue refined_value;
-        if (interpolation_refine == HdInterpolationVertex) {
-            refined_value = refiner->RefineVertexData(token, role, values[i]);
-        } else if (interpolation_refine == HdInterpolationFaceVarying) {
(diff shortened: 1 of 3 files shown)


plugin/hdCycles/mesh.cpp
```diff
@@ -653,6 +672,87 @@ HdCyclesMesh::_PopulateMotion(HdSceneDelegate* sceneDelegate, const SdfPath& id)
...
+    const HdCyclesMeshRefiner* refiner = m_topology->GetRefiner();
+
+    attr_m = attributes->add((ccl::AttributeStandard)cycles_motion_attribute);
+    ccl::float3* m = attr_m->data_float3();
+
+    for (unsigned int i = 0; i < numSamples; ++i) {
+        if (times[i] == 0.0f)  // todo: more flexible check?
+            continue;
+
+        VtValue refined_value;
```
bareya, line 715: Wouldn't [this](https://github.com/tangent-opensource/hdBlackbird/blob/main/plugin/hdCycles/meshRefiner.h#L102) work?

dedoardo: Oh yes, didn't know it existed. Thanks for pointing it out

skwerner (approved):

PR #214, opened 2021-06-07
Bugfix/fix warnings

skwerner:
This reverts a change that broke builds on Windows.

### DIFF
--- a/cmake/FindHoudiniUsd.cmake
+++ b/cmake/FindHoudiniUsd.cmake
@@ -97,14 +97,20 @@ foreach(_pxr_lib ${_houdini_pxr_libs})
 endforeach()
 
 # Find Usd Schema Generator
+
+if(APPLE)
+        set(HOUDINI_BIN ${HOUDINI_ROOT}/Resources/bin)
+else()
+        set(HOUDINI_BIN ${HOUDINI_ROOT}/bin)
+endif()
+
 if (NOT USD_SCHEMA_GENERATOR)
         find_program(USD_SCHEMA_GENERATOR
                 NAMES
                 usdGenSchema.py
                 usdGenSchema
                 PATHS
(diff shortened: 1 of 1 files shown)


cmake/FindHoudiniUsd.cmake
```diff
@@ -98,14 +98,20 @@ foreach(_pxr_lib ${_houdini_pxr_libs})
 endforeach()
 
 # Find Usd Schema Generator
+
+if(APPLE)
```
bareya, line 102: I still don't understand why we need this variable and can't we use `find_program` as simple as:

```cmake
        find_program(USD_SCHEMA_GENERATOR
                NAMES
                usdGenSchema.py
                usdGenSchema
                PATHS
                ${HOUDINI_ROOT}/bin
                ${HOUDINI_ROOT}/Resources/bin
                REQUIRED
                NO_DEFAULT_PATH
                )
```

In general I am against introducing redundant variables.

skwerner: The same path is being needed right after that:
`list(PREPEND USD_SCHEMA_GENERATOR ${HOUDINI_BIN}/hython)`

or should we run find_program() for python too?

dedoardo: I also don't see how to get around defining a variable. Maybe HOUDINI_BIN is more likely to be used somewhere else than defining one for HOUDINI_HYTHON or HOUDINI_SEARCH_PATHS. If the problem really is possible collisions we can just use different names.
Either way, I would like for this to be merged :)

dedoardo (approved):

sergeneren (approved):

bareya:
@dedoardo to avoid if statements and using that variable is easy as using `get_filename_component`.

PR #215, opened 2021-06-07
Overscan

dedoardo:
This patch handles values of `dataWindowNDC` which are outside the 0-1 range and thus produce overscan. [This core patch](https://github.com/tangent-opensource/coreBlackbird/pull/62) is required.

The idea is to scale all corners by the maximum amount of overscan so that the camera sensor can also be scaled uniformly but then only render to the data window.

This is the current behavior with `dataWindowNDC = (-0.5, -0.5, 1.5, 1.5)`. From my experiments, it always seems to match Karma (behavior-wise in Houdini)

**Husk to exr file** 
The .exr metadata reports 
`'dataWindow': (-1024, -540) - (3071, 1619)`
`'displayWindow':(0, 0) - (2047, 1079)`

**Viewport (Fits data window to viewport)**
viewport

**Husk to mplay (Shows display window)** 
mplay

**No overscan**
original

### DIFF
--- a/plugin/hdCycles/config.cpp
+++ b/plugin/hdCycles/config.cpp
@@ -121,7 +121,7 @@ HdCyclesConfig::HdCyclesConfig()
     shading_system = HdCyclesEnvValue<std::string>("HD_CYCLES_SHADING_SYSTEM", "SVM");
     display_buffer_linear = HdCyclesEnvValue<bool>("HD_CYCLES_DISPLAY_BUFFER_LINEAR", true);
 
-    max_samples = HdCyclesEnvValue<int>("HD_CYCLES_MAX_SAMPLES", 16);
+    max_samples = HdCyclesEnvValue<int>("HD_CYCLES_MAX_SAMPLES", 512);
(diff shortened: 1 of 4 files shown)


dedoardo:
This has been tested by me and @boberfly and seems to be working. DCCs are able to read the overscan data rendered through husk and renderbuffers in Houdini and mplay behave like Karma.

boberfly:
Hey @dedoardo 
I haven't been able to test this thoroughly. Will the projection matrix for the cycles-end work correctly if the overscan dimensions are not the same? We may want to have more in vertical than horizontal, or have more to the left than the right or up/down. I saw that overscan was just 1 float value so will this be an issue?

Cheers

dedoardo:
@boberfly No issue with different overscan amounts per border. The single overscan value (the maximum) is used to scale the render buffers but only a subrect is being rendered and that can be arbitrary.

skwerner (approved):

bareya (approved):

PR #217, opened 2021-06-10
Improvements to transformation motion blur:

skwerner:
* Number of steps does not need to be odd any more
 * No motion blur sent to coreBlackbird when all steps are the same


skwerner:
Resolving Freshdesk 7991

sergeneren (approved):

PR #221, opened 2021-06-14
Improved instancing performance

skwerner:
Instead of adding/deleting instances one by one, they're now allocated,
added, freed and removed in blocks. This makes it possible to use
millions of instances in an interactive session.


skwerner:
After this, the bottleneck is in coreBlackbirds' object surface area calculation. That can be skipped for renders that don't use OSL, because that property is not used outside of it.

bareya (approved):

bareya:
@skwerner does this solve problem #218 ?

skwerner:
I'd probably drop #128 after this one, yes.

PR #222, opened 2021-06-17
Bugfix/ortho cameras

skwerner:
Fixing ticket CD-131

Ortho cameras now get the correct transforms.

### DIFF
--- a/plugin/hdCycles/camera.cpp
+++ b/plugin/hdCycles/camera.cpp
@@ -381,7 +381,6 @@ HdCyclesCamera::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* renderParam,
         // flipping the Z axis.
 
         sceneDelegate->SampleTransform(id, &m_transformSamples);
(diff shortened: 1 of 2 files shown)


bareya (changes requested):
Hi @skwerner I think this PR requires some more work, ortho cameras are still busted:

Screenshot from 2021-07-08 08-30-10
Screenshot from 2021-07-08 08-30-28

skwerner:
@bareya I believe there is a Houdini/Hydra miscommunication in between there. Houdini GL is showing something different than all the Hydra delegates (Karma, Storm, Blackbird). In my test, Blackbird and Karma are giving matching results.

dedoardo:
I noticed that if you change resolution with a render settings node in Houdini, the image becomes stretched. This seems to be related to some missing updates to the camera because if the resolution is set to the older one, then the stretching disappears and the image matches karma.
Karma (Render settings 2048x1024)
image
Blackbird (Render settings 2048x1024)
image

PR #229, opened 2021-06-30
Wrap common settings under HdBbRPrim

bareya:
Remove code duplication and slide new `HdBbRPrim` class under mesh, points and curves. `HdBbRPrim` carries helper methods to manage object properties.


boberfly:
Hey @bareya this is great! The de-duplication of code will make it less error-prone and it looks like xforms can have visibility flags with their children inheriting those values correctly which didn't work before.

Instancing/prototypes seems to not work yet (crag asset doesn't have the flags working) but this patch didn't break anything, just something that's missing still that didn't work before this patch anyways.

PR #250, opened 2021-07-20
Instancer primvar

dedoardo:
**Description**
This patch forward primvars authored on a point instancer to a additional set of attributes which takes precedence over the geometry's attribute set and can be indexed by each object.
So far we are only forwarding attributes with per-instance interpolation as HdInterpolationInstance seems to be the only one available. Since the core patch supports any interpolation, is it possible to extract primvars with different interpolation from the instancer in Hydra? If I am reading correctly, [the USD documentation for point instancer](https://graphics.pixar.com/usd/docs/api/class_usd_geom_point_instancer.html#details) mentions that it's possible for primvars to have different interpolation.

It is the glue code for this [coreBlackbird PR](https://github.com/tangent-opensource/coreBlackbird/pull/91)

**Changes**
- Registering attribute sources for the instanced mesh attribute set 
- Registering the instanced attribute sets in the scene

**Todo**
Doing the same thing for volumes is straightforward, but I am investigating primvar overrides to figure out if they should go through an InstanceGroup.

### DIFF
--- a/plugin/hdCycles/attributeSource.cpp
+++ b/plugin/hdCycles/attributeSource.cpp
@@ -380,8 +380,6 @@ HdBbAttributeSource::ResolveAsArray()
 
     assert(num_src_comp <= num_dst_comp);
 
-    std::cout << "Committing " << name << " " << src_size << std::endl;
-
     // if Cast fails we must recover
     if (!src_data || !dst_data) {
         return false;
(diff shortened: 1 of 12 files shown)


dedoardo:
I think we should skip creating an instance group if there are no primvars to be sourced.
Maybe if the instance group has size one we should merge the attributes with the geometry attribute set?

bareya:
@dedoardo after I am done with building houdini artifacts I am going to have a look into this PR :)

PR #253, opened 2021-07-22
Sourcing constant primvars in volumes

dedoardo:
**Description**
The volume synchronization code does not read primvars besides common object setting. This code reads constant primvars into the geometry attributes.

For the primvars to be used in attribute nodes and then plugged to volume shaders, you need [this core patch](https://github.com/tangent-opensource/coreBlackbird/pull/96).

FD-8574

**Changes**
- Added object attribute source for constant primvars in the volume primitive

**Test**
Here is a test file with "customColor" and "customDensity" connected to a principled volume node
[test-volume-constant-primvars.zip](https://github.com/tangent-opensource/hdBlackbird/files/6864364/test-volume-constant-primvars.zip)


bareya (approved):
Looks good, you fixed long neglected volumes :)

sergeneren (approved):

PR #255, opened 2021-07-23
Better reading of Blender color spaces

skwerner:
Blender->USD appears to write the "Filmic sRGB" space as "Filmic_sRGB".
Added special cases to change those color spaces back to our names.

### DIFF
--- a/plugin/hdCycles/material.cpp
+++ b/plugin/hdCycles/material.cpp
@@ -484,6 +484,15 @@ convertCyclesNode(HdMaterialNode& usd_node, ccl::ShaderGraph* cycles_shader_grap
     if (cycles_node_name == "image_texture") {
         ccl::ImageTextureNode* tex = static_cast<ccl::ImageTextureNode*>(cyclesNode);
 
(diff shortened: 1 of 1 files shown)


plugin/hdCycles/material.cpp
```diff
@@ -484,6 +484,15 @@ convertCyclesNode(HdMaterialNode& usd_node, ccl::ShaderGraph* cycles_shader_grap
     if (cycles_node_name == "image_texture") {
         ccl::ImageTextureNode* tex = static_cast<ccl::ImageTextureNode*>(cyclesNode);
 
+        // Tangent Animation specific fix - Blender->USD inserts underscores in those color spaces.
+        if (tex->colorspace == "Filmic_sRGB") {
```
bareya, line 488: Why some of them have underscore some of them don't? :)

skwerner: Because it's late and I'm half asleep already :)

bareya (approved):

